PR #223, opened 2015-02-23
Fix argument list error in accept_ma

eirikgje:


lpsinger:
All of the Anaconda builds in Travis are failing, though this appears to be unrelated. I triggered a rebuild of master to check.

lpsinger:
Travis build issue was transient. @zonca, good to merge, in my opinion :+1:

zonca:
@eirikgje what about `smoothing`? there the first argument is `maps`

eirikgje:
Hm, so it shouldn't pose a problem, then? Not sure what you mean @zonca ?

zonca:
if a user calls `smoothing(maps=)`

eirikgje:
Ah, sorry, now I understand. You prefer to avoid having a 'unified' interface so that it will always be map_in? If not, I will just change the smoothing function. Otherwise, maybe it's not possible to fix this with the way it is set up now?

zonca:
I think it is fine if we change smoothing input to be map_in, can you add
that to the PR?

eirikgje:
Yup, sure

lpsinger:
> I think it is fine if we change smoothing input to be map_in, can you add
> that to the PR?

Does `smoothing` take `maps` currently because it can handle a vector-valued HEALPix map? Can all of these other functions do that too? If so, this sounds fine.

eirikgje:
Another solution: To decorate accept_ma with another decorator, so that the accept_ma function can be 'aware' of what the name of the 'map' argument is supposed to be. I suppose it's more flexible, but if it's not flexibility that is really useful anyway, just changing smoothing is easier.

zonca:
@lpsinger all functions accept either 1 map or an IQU map. @eirikgje that is evil! let's just change `smoothing`.

thanks!

eirikgje:
Allright, I'll do it now then!

lpsinger:
:+1: good to merge

PR #228, opened 2015-03-11
Replace xrange with six.moves.xrange to support Python 3

### LINKED ISSUE #226: Python 3 incompatibility in projaxes._make_segment() - xrange undefined
heapy 1.8.4 as installed with pip has a Python 3 incompatibility in projaxes._make_segment().  Line 397 of projaxes.py calls the Python 2 xrange() builtin function but this function is not present in Python 3.  In Python 3 the Python 2 xrange() have been renamed range() and the Python 2 range() has been removed.

An ugly but easy fix would be to add
```Python
try:
    xrange
except NameError:
    xrange = range
```
in the module preamble.

lpsinger:
As we have already done in a few other places. Also add missing dependency on six. Fixes #226.

### DIFF
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -24,6 +24,7 @@ try:
 except ImportError:
     import pyfits as pf
 import numpy as np
+import six
 from . import pixelfunc
 from .sphtfunc import Alm
 import warnings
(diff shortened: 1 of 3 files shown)


lpsinger:
tests passed, good to merge :+1:

PR #230, opened 2015-03-16
Use integer division in visualization functions

### LINKED ISSUE #227: Python 3 incompatibilities in visufunc - sub relies on implicit integer division
The subplot functionality of the visualisation functions in visifunc (mollview(), gnomview(), cartview(), orthview()) does not work correctly with Python 3 because it relies on implicit integer division.  The problematic lines are `c,r = (idx-1)%ncols,(idx-1)/ncols` which calculate the column and row number for the plot.  Under Python 3 `(idx-1)/ncols` will do 'true' division and return fractional values, causing plots in columns after the first to be offset below their correct position.  For compatibility with both Python 2 and 3 explicit integer division should be used, i.e. replace each instance of `(idx-1)/ncols` with `(idx-1)//ncols`

lpsinger:
Fixes #227.

### DIFF
--- a/healpy/visufunc.py
+++ b/healpy/visufunc.py
@@ -158,7 +158,7 @@ def mollview(map=None,fig=None,rot=None,coord=None,unit='',
             raise ValueError('Wrong values for sub: %d, %d, %d'%(nrows,
                                                                  ncols,
                                                                  idx))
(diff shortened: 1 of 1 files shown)


zonca:
thanks!

PR #231, opened 2015-03-20
write_alm can store a list of alms

arahlin:
This code has been tested and produces an output that is compatible with fits2alm.


zonca:
Great! thank you very much for your contribution!

PR #232, opened 2015-03-21
LIGO enhancements to write_map

lpsinger:
This is a batch of enhancements for `write_map` and some backward compatibility patches, all related to writing LIGO probability sky maps:

* support adding units to the FITS header for each column in the HEALPix map
* support writing gzip-compressed FITS files, even with old versions of pyfits
* some minor changes in numpy and pyfits usage to support older versions of these packages that come with Scientific Linux 6, the OS of several of the LIGO computing clusters


zonca:
Thanks @lpsinger !

PR #233, opened 2015-03-23
Fixed build on readthedocs

zonca:
finally got around to fix this.

@lpsinger I had to set `setuptools==12.0.5` in `requirements.txt`.
Do you think this could have implications for the install process?

PR #234, opened 2015-03-24
Update to latest version of setuptools bootstrap script

lpsinger:
@zonca, does this fix #233?


zonca:
Easier if I do this on that same branch, which is already setup on readthedocs!

PR #235, opened 2015-03-24
Add argument to write_map for extra header fields

lpsinger:
I forgot to add this to #232.


lpsinger:
@zonca, tests pass. What do you think?

zonca:
@lpsinger thanks, this is also useful

PR #244, opened 2015-04-23
Add missing top-level import for get_interp_weights

### LINKED ISSUE #240: rename "get_neighbours" and adjust documentation
The function "get_neighbours" in pixelfunc.py should be renamed to something like "get_interpol" (which is the name of the corresponding function in the C++ library.
The documentation should be changed from
    """Return the 4 nearest pixels and corresponding weights.
to something like
  """Return the 4 closest pixels on the two rings above and below the location and corresponding weights. Biinear interpolation is carried out along latitude and longitude.

lpsinger:
Fixes #240.


zonca:
@lpsinger released 1.8.6 to fix this

PR #249, opened 2015-04-30
Allow read_map to also take an HDUList or HDU instance

astrofrog:
Not tested yet since I'm having issues compiling healpy on my machine, but hopefully Travis can do the testing :smile:

### DIFF
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -453,7 +453,7 @@ def _get_hdu(input_data, hdu=None, memmap=None):
             raise ValueError('Available hdu in [0-%d]' % len(hdulist))
         else:
             fits_hdu = input_data[hdu]
-    elif isinstance(input_data, (pf.PrimaryHDU, pf.ImageHDU)):
+    elif isinstance(input_data, (pf.PrimaryHDU, pf.ImageHDU, pf.BinTableHDU, pf.TableHDU, pf.GroupsHDU)):
         fits_hdu = input_data
     else:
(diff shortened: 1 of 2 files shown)


lpsinger:
Nice! Could you do the same for `read_cl`, `read_alm`, `write_map`, `write_cl`, and `write_alm`?

astrofrog:
@lpsinger - I implemented this for the read functions using a helper function. I don't think it makes sense for the write functions though? I also did some minor PEP8 cleanup.

lpsinger:
> I don't think it makes sense for the write functions though?

Why not?

astrofrog:
@lpsinger - so I guess you mean that if I passed an HDU, it would write the data to that HDU? Or that the ``write`` functions could either write to a file or return an HDU?

lpsinger:
> @lpsinger - so I guess you mean that if I passed an HDU, it would write the data to that HDU? Or that the write functions could either write to a file or return an HDU?

Oh, I see the issue... maybe we just want to factor the writing into a function that creates and HDU and a function that calls that and then writes it to a FITS file.

zonca:
Thanks @astrofrog,
@lpsinger fine with me.

astrofrog:
@lpsinger - unit tests added

lpsinger:
Nice! :+1: to merge.

zonca:
thanks

lpsinger:
Hey @zonca. Do you have any releases coming up? I'd like to use this feature in one of my projects.

zonca:
yes, I need also one of the bugfixes in mollview. I'd release a new version
next week.

Issue #252, opened 2015-05-06
Add utility functions to work with "cutouts"


cdeil:
Disclaimer – I know almost nothing about HEALPIX ... apologies if this doesn't make sense.

My understanding of the HEALPIX <-> normal WCS projection image reprojection functions in https://github.com/astrofrog/reproject is that the HEALPIX data will always cover the whole sky, i.e. there'll be lots of zeros if the image only covers a small part of the sky.

Would it make sense to add functionality to `healpy` or `reproject` or `pymoc` to efficiently handle this case, i.e. store only the pixel values within a footprint in the HEALPIX data array and FITS file when reprojecting a normal image to HEALPIX?

Another use case would be to produce a cutout of an all-sky HEALPIX image ... I guess this is what the web server does here http://aladin.u-strasbg.fr/hips/ to only send the data to the user that's in the viewport, but it is not easy to do this cutout with `healpy`?

cc @astrofrog as the https://github.com/astrofrog/reproject dev
cc @tboch as the https://github.com/tboch/pymoc dev

zonca:
`healpy` has partial support for explicit pixelization maps, i.e. dictionary where keys are pixel numbers.
see in the projector:

https://github.com/healpy/healpy/blob/a6ae7de250c937b3724985eef4f429502ca90d3f/healpy/projector.py#L185

However it doesn't allow to use the most powerful feature of HEALPix which is the spherical armonic transforms. I never used it.
It would be interesting if you could give it a try to check if it still works.

zonca:
See #254, it would be useful to check if the visualization functions still work with explicit pixelization maps.

It is a corner use case, I won't spend time on this.

Closing for now, please reopen if you want to contribute some test results.

PR #254, opened 2015-05-08
write_map and read_map support for partial sky maps

arahlin:
Added support for reading and writing partial sky maps.

New keywords OBJECT and GRAIN are written by write_map, and an index column is added if partial=True.  read_map parses INDXSCHM, OBJECT and GRAIN to determine whether the map
is a full or partial sky map, and returns an UNSEEN-padded map if the input map is partial.  The field index begins from the first column after the pixel index.

This might be useful for #252 ?

### DIFF
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -226,10 +226,6 @@ def write_map(filename,m,nest=False,dtype=np.float32,fits_IDL=True,coord=None,pa
                         'Indexing: IMPLICIT or EXPLICIT')
     tbhdu.header.update('OBJECT', 'PARTIAL' if partial else 'FULLSKY',
                         'Sky coverage, either FULLSKY or PARTIAL')
-    tbhdu.header.update('GRAIN', 1 if partial else 0,
-                        'GRAIN = 0: No index')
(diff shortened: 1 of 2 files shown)


zonca:
thanks @arahlin ! what is the meaning of the `GRAIN` keyword?

arahlin:
In trying to make this method sort of compatible with typical uses of partial-sky maps, I ran across this keyword.  I believe that a value of GRAIN > 1 mean that one pixel index corresponds to that many consecutive map pixels.  This way, you can control the size of the stored map, the indexing scheme, and the granularity with separate keywords.  In principle, support for that can be added in, but as it stands I guess the keyword is kind of redundant.

zonca:
I would remove the GRAIN keyword, apart from that the patch looks good to merge. maybe you can print out a warning if that keyword is present in the file.

arahlin:
Sure I can do that.

arahlin:
I didn't bother adding a warning, the GRAIN keyword seems uncommon enough that it's unlikely to be used by healpy users.

zonca:
great, thanks!

PR #258, opened 2015-05-28
option to read all fields with read_map

arahlin:
If the field argument is explicitly None, read all the columns from the fits file.  The default behavior is still to read in the first field in the file.


zonca:
thanks,
can you fix it for Python 3?

it complaints about `basestring`:
https://travis-ci.org/healpy/healpy/jobs/64413430

arahlin:
is this ok to merge?

zonca:
oh yes ,thanks

PR #259, opened 2015-05-29
Fix typo and missing keywords in the handling of a set of spin-0 maps.

pchanial:
In the spin-0 case, the sequence of maps is iterated but the element is not used. In addition, the lmax and mmax keywords are not propagated (and the mask stuff doesn't need to be iterated).


zonca:
thanks!

PR #261, opened 2015-06-04
Remove Cython & numpy dependencies when querying setup.py

evertrol:
When installing through pip and using the requirements file, pip makes two passes through the `setup.py` files of the packages (including healpy): the first pass is to obtain some info, through the `setup.py egg_info` command. 
In a clean environment, where numpy and Cython are not yet installed (e.g., a virtualenv), this result in an error, since `setup.py` unconditionally imports numpy and Cython. Using a conditional import, by testing the argument given to `setup.py`, allows one to install healpy in one go (`pip install healpy`) without first needing to install numpy and Cython separately. This also prevents problems when one runs e.g. `setup.py --help` in a clean environment.
More information can be found at the [pip issue tracker](https://github.com/pypa/pip/issues/25); the solution given there at the bottom, is the one implemented here.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,10 +86,16 @@ on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 
 cython_require = 'Cython >= 0.16'
 try:
-    pkg_resources.require(cython_require)
-    from Cython.Distutils import build_ext
-    ext = "pyx"
-    extcpp = "pyx"
+    if ('--help' in sys.argv[1:] or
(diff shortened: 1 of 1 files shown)


zonca:
@lpsinger could you please review and merge this?

zonca:
@lpsinger can you take a look at this? You have a better understanding of the install process!

zonca:
looks fine to me, thanks @evertrol

Issue #262, opened 2015-06-24
ud_grade modifies input array


lpsinger:
The `ud_grade` function modifies the input array in some circumstances. This is probably an undesired side effect. The issue occurs if there are any invalid values (`hp.UNSEEN`) in the input array. In this case, the corresponding entries are overwritten with zero in `_ud_grade_core`:

        m[bads] = 0

zonca:
can you provide an example that fails?

`UNSEEN` should be restored back at https://github.com/healpy/healpy/blob/master/healpy/pixelfunc.py#L1664

lpsinger:
Oh, I didn't notice that. Still, it's generally bad practice to modify the input. This will not work for the most general Numpy array. See, for instance, the following example:

```python
>>> import numpy as np
>>> import healpy as hp
>>> m = np.repeat(hp.UNSEEN, 48)
>>> m.flags.writeable = False
>>> hp.ud_grade(m, 1, order_in='NEST', order_out='NEST')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/healpy/pixelfunc.py", line 226, in wrapper
    out = f(m, *args, **kwds)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/healpy/pixelfunc.py", line 1613, in ud_grade
    mout = _ud_grade_core(m,nside_out,pess=pess, power=power, dtype=dtype)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/healpy/pixelfunc.py", line 1650, in _ud_grade_core
    m[bads] = 0
ValueError: assignment destination is read-only
```

zonca:
Any other option without making a copy of the map?

lpsinger:
Not that I can think of.

cjcopi:
Replacing https://github.com/healpy/healpy/blob/master/healpy/pixelfunc.py#L1653 with
```
map_out = np.sum(mr*hit, axis=1).astype(type_out)
```
and removing the writes to `m` would make the function work for read only maps.

zonca:
Great! Could you create a pull request for that?

zonca:
fixed in #264

PR #264, opened 2015-07-02
Allow ud_grade to work with read only maps.

cjcopi:
Work around for issue #262.

### DIFF
--- a/healpy/pixelfunc.py
+++ b/healpy/pixelfunc.py
@@ -1639,18 +1639,15 @@ def _ud_grade_core(m,nside_out,pess=False,power=None, dtype=None):
         ratio = 1
     
     if nside_out > nside_in:
-        rat2 = npix_out/npix_in
+        rat2 = npix_out//npix_in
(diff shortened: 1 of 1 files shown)


healpy/pixelfunc.py
```diff
@@ -1647,10 +1647,9 @@ def _ud_grade_core(m,nside_out,pess=False,power=None, dtype=None):
         bads = (mask_bad(m) | (~np.isfinite(m)))
         hit = np.ones(npix_in,dtype=np.int16)
         hit[bads] = 0
```
lpsinger: Note that `hit` is equivalent to `~bads`.

cjcopi:
Updated to remove hit.  I changed bads to goods since that is what is actually wanted and it can be calculated from the reshaped array (mr).  Since rat2 is an integer I also enforce integer division.  The logic of the goods assignment could be "multiplied out" but I have not done so.

lpsinger:
:+1:

zonca:
great thanks!

Issue #163, opened 2014-02-24
Add nside2order, order2nside


lpsinger:
Could `order2nside` and `nside2order` methods be added? The former is trivial (it's just `1<<order`) but implementing the latter in exact integer arithmetic is nontrivial and there is a slick implementation in `healpix_base.cc`.

cjcopi:
This seems to have fallen through the cracks, but is simple to implement so might as well be done.  As noted `order2nside` is just a bit shift operation.  Similarly `nside2order` is just a logarithm base 2.  Implementing this by converting `log2` to an integer is not ideal so there seems to be at least two options

1. Use `frexp` as in `np.frexp(nside)[1]-1`,
2. Use `ilog2`.  A nice implementation of this is in Healpix math_utils.h which can be directly converted to Python.

Option 1 has the side effect of working with floats and arrays automatically.  `ilog2` is based on bit shifting so would require more work to vectorize.

Is there a preferred implementation?

zonca:
I prefer 1, better to have less code to maintain

mreineck:
On 07/07/15 23:45, Andrea Zonca wrote:
> I prefer 1, better to have less code to maintain
>
But please be careful about rounding errors with frexp! For very large
integers there is no exact floating point representation, and you might
get wrong results.

Cheers,
  Martin

cjcopi:
> But please be careful about rounding errors with frexp! For very large integers there is no exact floating point representation, and you might get wrong results.

This is a general concern with using floating point routines, but is it a concern here?  Do you have an example?  Without having looked at the code for frexp I was under the impression it was safer in this regard.  This is particularly true given that nside is a power of 2.

In testing (only on 64 bit Linux) `np.frexp` returns the correct answer up to `order=63` after which it throws an error due to conversion issues.  Interestingly `math.frexp` works up to `order=1023` before it throws an error.

At present in healpy `check_nside` restricts the order to less than 30 so perhaps this is a moot point.

mreineck:
The potential rounding is not a problem in this concrete situation. I just wanted to warn about replacing the general-purpose ilog2 with a frexp-based approach. As long as the frexp call is performed on numbers known to be powers of 2, everything is fine.
However if healpy should ever be enhanced to support non-power-of-2 values for Nside (in RING scheme only, of course), then the nside2order function will have to return -1 for such inputs in order to be analogous to the C++ version.

lpsinger:
[...]

PR #266, opened 2015-07-08
Implement nside2order and order2nside.

cjcopi:
Initial implementation of nside2order and order2nside as discussed in issue #163.  Both functions have been imported into the healpy namespace.


zonca:
@cjcopi it fails on Python 2.6, there is no `bit_length` method there, is it possible to fix this? I think we prefer to still support Python 2.6 ( @lpsinger can comment more on this)

lpsinger:
We don't use Python 2.6 on the LIGO clusters anymore. So Python 2.6 is up to you (though we should remove some existing workaround code if we drop Python 2.6 support).

zonca:
thanks @lpsinger, then I would drop support for python 2.6, I'll merge this as it is and open another issue.

PR #268, opened 2015-07-10
Drop support for Python 2.6

### LINKED ISSUE #267: Drop support for Python 2.6
Almost nobody uses Python 2.6 anymore, we should drop support for it, so we can cleanup some additional (ugly) code that was required for that (mainly in `setup.py`).

We should also drop the Python 2.6 tests from the travis-CI configuration.

Anybody volunteers to do this?

lpsinger:
Fixes #267.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@
 language: python
 
 python:
-  - 2.6
   - 2.7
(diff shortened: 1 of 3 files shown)


zonca:
thanks!

PR #271, opened 2015-07-16
return a result when boundaries() is called with the healpixels given as python list

segasai:
Hi, 
currently when boundaries() method is called like this 

healpy.boundaries(2,[2,3]) 
it returns None 

This pull requests fixes this by trying to cast the input to a numpy array.

Cheers,
        Sergey


lpsinger:
Nice bugfix. I would suggest revamping the type checking to use `np.isscalar` and `np.can_cast`, as those functions have sorted out all the corner cases of different numeric and collection types.

segasai:
Okay, I'll take a look at np.can_cast & np.isscalar

zonca:
@segasai, do you plan to work more on this? Thanks!

zonca:
moved to #279

PR #276, opened 2015-09-03
Python 3 compatibility: alias long to int

evertrol:
This prevents an exception at the line 

    bins = long(min(data2.size/20, 5000))


lpsinger:
I'm not sure why the cast to `long` was necessary in the first place. How about just changing this line to use true division (`//`) and no cast?

        bins = min(data2.size//20, 5000)

evertrol:
I've changed the PR to use integer divison instead.

zonca:
Thanks!

lpsinger:
:+1:

PR #277, opened 2015-09-04
A few minor fixes for write_map and write_alm

arahlin:
I've found the following two minor updates useful:

 * raise a useful `ValueError` for empty partial map in `write_map`
 * test alms with `cookbook.is_seq_of_seq` (instead of `isinstance`) in `write_alm`


zonca:
Great thanks!

PR #281, opened 2015-09-17
TEST: using travis new container infrastructure

zonca:
cannot use sudo, using addons apt instead

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,44 +4,67 @@
 
 language: python
 
+sudo: false
+
+addons:
+    apt:
+        packages:
+        - pkg-config
+
 python:
   - 2.7
   - 3.3
   - 3.4
 
 env:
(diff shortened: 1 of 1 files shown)


lpsinger:
I think it is still important to have tests that use `pip`.

zonca:
@lpsinger sure, trying to keep all the same tests

zonca:
@lpsinger I think this is pretty good.

however I cannot implement the removal of packages:

    sudo apt-get -y remove autoconf automake pkg-config && \

do you think this is important?

lpsinger:
> do you think this is important?

Yes, it's important. Those packages need to be removed because we are testing that the package builds without a full autotools developer installation.

zonca:
ok, asked to the support of travis-ci

zonca:
travis-ci confirmed it is not possible to remove packages in the new infrastructure, so let's keep using the legacy infrastructure for now.

PR #282, opened 2015-09-18
Change header.update to header.set or dict format

eirikgje:
AstroPy throws a DeprecationWarning when using header.update. This patch
updates fitsfunc so that it uses header.set or the header[key] = value
format instead

### DIFF
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -116,7 +116,7 @@ def write_cl(filename, cl, dtype=np.float64):
 
     tbhdu = pf.new_table(cols)
     # add needed keywords
-    tbhdu.header.update('CREATOR','healpy')
+    tbhdu.header['CREATOR'] = 'healpy'
     writeto(tbhdu, filename)
 
 def write_map(filename,m,nest=False,dtype=np.float32,fits_IDL=True,coord=None,partial=False,column_names=None,column_units=None,extra_header=()):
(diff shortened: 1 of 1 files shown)


mreineck:
Funny ... that single Travis failure looks spurious to me. Is there a way to re-run the tests?

zonca:
Thanks!

lpsinger:
May I check that this is compatible with our LIGO computing clusters? That was why we had the antiquated pyfits stuff.

zonca:
@lpsinger ok, I'll wait for your answer before releasing

zonca:
@lpsinger is this compatible with LIGO clusters?

lpsinger:
No, this doesn't work with our Debian clusters, which use PyFITS 3.0.8.

Also, our Scientific Linux clusters are still running Python 2.6, so we might have to roll back #268.

lpsinger:
Works fine on LIGO clusters if I revert this and #268. Shall I open a PR to do that?

mreineck:
Just a heads-up related to this backwards-compatibility discussion: I plan to require a C++11-capable compiler for future releases of Healpix C++ (after 3.30), and I doubt that the gcc on the LIGO installations will be modern enough for this...

zonca:
Ok Leo, do you know how long will we need to support this generation of
LIGO clusters?

lpsinger:
> Just a heads-up related to this backwards-compatibility discussion: I plan to require a C++11-capable compiler for future releases of Healpix C++ (after 3.30), and I doubt that the gcc on the LIGO installations will be modern enough for this...

C++0x has been supported by gcc for a long time, but C++11 was introduced in gcc 4.7. Our Debian clusters have gcc 4.7.2 and support C++11, but our Scientific Linux 6 clusters have gcc 4.4.5 and do not support it.

My best guess is that we will probably upgrade to Scientific Linux 7, bringing Python 2.7 and C++11, by January, shortly after the end of our first observing run.

mreineck:
C++0x and C++11 are different names for the same thing. gcc has been tracking the evolution of the standard aggressively, but the support is incomplete up to version 4.8.1, and even there some library features are missing.

I don't plan to use new features gratuitously, but requiring a compiler that is not older than, say, 3 years seems reasonable to me.

At the same time, older versions of Healpix C++ will still receive bug fixes, so nobody will be left abandoned with a buggy Healpix library on an older platform; that's a promise!

Issue #286, opened 2015-10-09
orthview projplot fix


rwogburn:
When plotting with projplot on top of an orthview plot, all single points show up on the left half-sky.  See code snippet below.  This can be fixed by changing projector.py line 744 from x\*=1 to x\*=-1.

Thanks,
Walt

  nside = 8
  npix = 12\*nside\*nside
  rot=[0,90,0]

  # Plot linear ramp of pixel values
  v = np.arange(npix)
  hp.orthview(v,rot=rot,nest=False,title="Circles left, squares right")

  # Plot filled circles at 45 degree N latitude
  hp.projplot(0,45,'ro',lonlat=True,rot=rot,markersize=10)
  hp.projplot(90,45,'go',lonlat=True,rot=rot,markersize=10)
  hp.projplot(180,45,'bo',lonlat=True,rot=rot,markersize=10)
  hp.projplot(270,45,'yo',lonlat=True,rot=rot,markersize=10)

  # Plot squares at 45 degree S latitude
  hp.projplot(0,-45,'rs',lonlat=True,rot=rot,markersize=7)
  hp.projplot(90,-45,'gs',lonlat=True,rot=rot,markersize=7)
  hp.projplot(180,-45,'bs',lonlat=True,rot=rot,markersize=7)
  hp.projplot(270,-45,'ys',lonlat=True,rot=rot,markersize=7)

zonca:
Fixed in #288

PR #288, opened 2015-10-12
RWO: Typo fix to fix hemisphere in orthview projplot

rwogburn:
Fixes issue #286 where points plotted with projplot on orthview appear in the wrong hemisphere.


zonca:
thanks!

PR #291, opened 2015-10-14
FIX: Switch submodule protocols from git to https

stratosgear:
Some of us are working behind company firewalls that
do not allow access to git:// protocols, thus making
it impossible to use these projects as is.

This commit changes to protocols from git:// to
https://.  New repository clones should work with the
new protocols just fine.

Existing repository clones will have to run:

    git submodule sync

in order to copy the protocol changes from
.gitmodules to .git/config


lpsinger:
:+1:

lpsinger:
@zonca, can you merge this?

Issue #287, opened 2015-10-09
gnomview does not honor 'degree' argument


rwogburn:
It's simply not used.  Can be fixed by adding following lines in visufunc.py somewhere before line 358:

```
if (degree):
  reso*=60
```

lpsinger:
I'd actually favor just dropping the unused argument. Kind of silly having two arguments just to specify one quantity, the angular resolution per pixel.

rwogburn:
Healpix gnomview doesn't have it, so I suppose it's fine for healpy not to have it either (as long as it's documented correctly).  Although you do end up with a mix of units since the rotation is specified in degrees.

rwogburn:
I have a topic branch to use the degree argument:
https://github.com/rwogburn/healpy/tree/issue287

and one to remove it instead:
https://github.com/rwogburn/healpy/tree/issue287alt

Either could be merged.  Thanks.

rwogburn:
Thanks lpsinger and zonca, this is resolved.

PR #294, opened 2015-10-15
Adjust resolution if gnomview degree argument set

rwogburn:
This causes the gnomview 'degree' argument to take effect (currently ignored).  Fixes issue #287.

If you prefer, the 'degree' argument could be removed instead (lpsinger's suggestion):
https://github.com/rwogburn/healpy/tree/issue287alt


zonca:
Thanks, better remove it , can you please do a PR with the other branch? Thanks!

PR #297, opened 2015-10-19
avoid C++11 features

mreineck:
This patch should remove problematic C++11 features from the C++ sources in healpy.
Similar features are present in the sources of Healpix 3.30 itself; I'm currently working to remove them as well.


zonca:
thanks @mreineck

Issue #295, opened 2015-10-15
Install issue on OS X 10.11


jisuoqing:
Hi,

I've tried to install healpy with a few ways, but when I imported the module, it says:

```
>>> import healpy
healpy/pixelfunc.py:105: UserWarning: Warning: cannot import _healpy_pixel_lib module
  warnings.warn('Warning: cannot import _healpy_pixel_lib module')
healpy/sphtfunc.py:34: UserWarning: Warning: cannot import _healpy_pixel_lib module
  warnings.warn('Warning: cannot import _healpy_pixel_lib module')
healpy/sphtfunc.py:38: UserWarning: Warning: cannot import _healpy_fitsio_lib module
  warnings.warn('Warning: cannot import _healpy_fitsio_lib module')
healpy/sphtfunc.py:42: UserWarning: Warning: cannot import _sphtools module
  warnings.warn('Warning: cannot import _sphtools module')
healpy/__init__.py:58: UserWarning: Warning: cannot import query disc module
  warnings.warn('Warning: cannot import query disc module')
healpy/__init__.py:62: UserWarning: Warning: cannot import pixelfunc module
  warnings.warn('Warning: cannot import pixelfunc module')
healpy/__init__.py:69: UserWarning: Warning: cannot import _sphtools module
  warnings.warn('Warning: cannot import _sphtools module')
healpy/__init__.py:77: UserWarning: Warning: cannot import pixel lib module
  warnings.warn('Warning: cannot import pixel lib module')
healpy/projaxes.py:36: UserWarning: Warning: cannot import _healpy_pixel_lib module
  warnings.warn('Warning: cannot import _healpy_pixel_lib module')
```

I've tried ``pip install healpy`` as well as the development install, but neither of them work.

Thanks!

lpsinger:
@zonca, I think that these conditional imports create a lot of confusion. It's better to stick to a single code path where possible. Why are they necessary? Does it have to do with readthedocs? There may be a cleaner solution.

zonca:
@lpsinger yes, readthedocs, at some point it was also possible to run only the pure python part of healpy (plotting) if you couldn't compile healpix c++, but haven;t tested that in a while.

lpsinger:
Does the readthedocs build environment have a working C/C++ compiler toolchain? If so, is it just a matter of it being slightly wasteful to build the native modules on readthedocs?

zonca:
It looks like still they do not support building C extensions

lpsinger:
[...]

PR #298, opened 2015-10-19
Remove special cases for imports and setup.py build phases

lpsinger:
There are numerous conditional imports and setup.py special cases to support building on Read The Docs. These tend to mask the causes of user installation errors (see #295).

The RTD build environment does support installing C extensions, so these patches cause RTD to do a full build of Healpy. RTD does lack autotools, which still must be downloaded as a special case.

The method `Distribution.fetch_build_eggs()` does the same thing as setting `setup_requires`, but at a controllable point in the script. The following changes are made to handling build dependencies:

* If the Cython-generated source files are absent, then Cython is required at all phases, including if the user is just asking for `--help`, because Cython adds command line options to `setup.py`. Because we are now using  `setuptools` to enforce this dependency, Cython will be downloaded as an egg if it is needed.
* Numpy is deferred to `build_ext`.
* `pkg-config` (or the fallback, `pykg-config`) is deferred to `build_clib`.
* As indicated above, we add a special case for Read The Docs, whose build environment lacks Autotools.

The Read The Docs output for this branch is here:
http://lpsingerhealpy.readthedocs.org/en/setup_cleanup/

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -12,48 +12,6 @@
 # serve to show the default value.
 
 import sys, os
-try:
-    from exceptions import ImportError
-except:
-    pass
-
-# To avoid problem with ReadTheDocs and compiled extensions.
-class Mock(object):
-    """Special Healpix values for masked pixels.
-    """
-    pi = 3.141516
-    class Axes(object):
-        pass
-    class Locator(object):
-        pass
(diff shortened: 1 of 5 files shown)


zonca:
good job @lpsinger that was a lot of refactoring!

PR #301, opened 2015-10-22
Ensure integer indices for python3 compatibility

pqrs6:


lpsinger:
Better off using integer division, `//`, here.

pqrs6:
Thanks for the heads up there, I didn't realize // worked for python2 as well as python3.

PR #302, opened 2015-10-22
Ensures subplot indices are integers

pqrs6:
Python3 compatibility issue, without this subplots are placed according to floating point division, not integer division as is implied in the script.


FriendsJessica:
How about the situation that subgraphs are difference in size?

PR #313, opened 2015-12-01
Enable custom dtypes for each map column for readmap and writemap

eirikgje:


zonca:
Thanks! Can you add a check that the maps you read back in are the same?

eirikgje:
Yup - I'll delete this PR and submit another

PR #338, opened 2016-04-04
Drop support for Python 2.6

lpsinger:


lpsinger:
@zonca, looks like this is ready to merge.

PR #341, opened 2016-05-18
Resolve deprecated method warning by pyFITS.

keskitalo:
Also fix non-integer division warning when running with Python 3.

### DIFF
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -20,6 +20,7 @@
 """Provides input and output functions for Healpix maps, alm, and cl.
 """
 from __future__ import with_statement
+from __future__ import division
 
 import six
 import gzip
@@ -208,7 +209,7 @@ def write_map(filename,m,nest=False,dtype=np.float32,fits_IDL=True,coord=None,pa
             mm2 = np.asarray(mm)
             cols.append(pf.Column(name=cn,
                                    format='1024%s' % curr_fitsformat,
-                                   array=mm2.reshape(mm2.size/1024,1024),
+                                   array=mm2.reshape(mm2.size//1024,1024),
                                    unit=cu))
         else:
             cols.append(pf.Column(name=cn,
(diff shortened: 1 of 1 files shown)


healpy/fitsfunc.py
```diff
@@ -208,15 +209,19 @@ def write_map(filename,m,nest=False,dtype=np.float32,fits_IDL=True,coord=None,pa
...
         else:
             cols.append(pf.Column(name=cn,
                                    format='%s' % curr_fitsformat,
                                    array=mm,
                                    unit=cu))
 
-    tbhdu = pf.new_table(cols)
+    try:
+        tbhdu = pf.BinTableHDU.from_columns( cols )
+    except:
```
lpsinger: Whoa. This is much to general. What exception are you trying to catch here?

healpy/fitsfunc.py
```diff
@@ -208,15 +209,19 @@ def write_map(filename,m,nest=False,dtype=np.float32,fits_IDL=True,coord=None,pa
...
                                    unit=cu))
         else:
             cols.append(pf.Column(name=cn,
                                    format='%s' % curr_fitsformat,
                                    array=mm,
                                    unit=cu))
 
-    tbhdu = pf.new_table(cols)
+    try:
+        tbhdu = pf.BinTableHDU.from_columns( cols )
```
lpsinger: What version of Astropy does this version depend on?

healpy/fitsfunc.py
keskitalo: Good point!

If the pyfits backend is old enough not to support the new API, it will
throw an AttributeError. I'll update the pull request.

healpy/fitsfunc.py
keskitalo: I get the deprecation warnings on the latest PyPI version of astropy: 1.1.2.

PR #342, opened 2016-05-20
Fix bug in write_alm

ziotom78:
Commit 29c7ff4777557754aa1cd163eb03d3a6133b5d03 removed `write_alm`'s ability to write multiple alms. This commit restores that functionality.


zonca:
THanks!

PR #343, opened 2016-05-30
Update projaxes.py

pqrs6:
Allows user to choose own default colormap so that cmap doesn't have to be specified in each call.


zonca:
Thanks, can you also explain how it works in the docstring?

pqrs6:
I expanded the comment right above the get_color_table() function, along with a brief example of how to set a colormap ahead of time.

zonca:
thanks!

Issue #340, opened 2016-05-11
Benign unresolved reference in `fitsfunc._get_hdu`


Goutte:
At https://github.com/healpy/healpy/blob/master/healpy/fitsfunc.py#L545 it should be `input_data` instead of `hdulist`.

I fixed that in my branch and can commit that along with my next PR (cf. issue #339), but if you'd rather have it in a different PR for git tree cleanliness (or even do it yourself) I'd understand.

zonca:
Yes please do a separate PR, thanks

zonca:
@Goutte any update on this?

Goutte:
Yes @zonca it is available in PR #353. There may be rebasing issues, tread carefully.

PR #353, opened 2016-09-20
Fix a benign unresolved reference.

Goutte:
See issue #340 .

I could not update the above issue itself with this PR, as I usually do with [hub](https://github.com/github/hub), possibly because I rebased ?

Anyhow, it's here.


zonca:
thanks!

Issue #307, opened 2015-10-28
Lambert azimuthal equal-area projection


rwogburn:
SPT shows maps in the [Lambert azimuthal equal-area projection](https://en.wikipedia.org/wiki/Lambert_azimuthal_equal-area_projection) (example: [Hanson et al. arxiv 1307.5830](http://arxiv.org/abs/1307.5830)).  It sometimes gets used by other experiments as well.  I have written a healpy projection and 'lambview' function in [rwogburn/healpy/tree/lambert](https://github.com/rwogburn/healpy/tree/lambert) to plot maps in this projection.  If it's useful for the main project I'll submit a pull request.

Here's an example, WMAP ILC plotted with 'lambview':
![wmap_lambert](https://cloud.githubusercontent.com/assets/15055831/10797839/cd289c20-7d61-11e5-8922-7a4a279148fb.png)

rwogburn:
The healpix IDL routines include an "azeqview" function for the [azimuthal equidistiant projection] (https://en.wikipedia.org/wiki/Azimuthal_equidistant_projection).  It should be straightforward to make the same projector class support both.

zonca:
forgot to answer this, yes, @rwogburn  I think a pull request would be useful!

rwogburn:
#354 :  This version adds one function 'azeqview' (as in Healpix IDL) with a flag to do equal-area instead of equidistant. Let me know if you prefer separate functions instead.

zonca:
was merged in https://github.com/healpy/healpy/pull/354

PR #354, opened 2016-09-21
 Lambert azimuthal equal-area projection #307 

rwogburn:
This version adds one function 'azeqview' (as in Healpix IDL) with a flag to do equal-area instead of equidistant.  Let me know if you prefer separate functions instead.


zonca:
Thank you very much for contributing this!

PR #365, opened 2016-11-21
Force ysize as int in np.arange and np.ones

evertrol:
I got a numpy warning when plotting with a MollWeide projection:
```
/usr/local/lib/python3.5/site-packages/numpy/core/numeric.py:190: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
```

I chased this down to `projector.py`: `ysize` was used directly in `np.arange` and `np.ones`, despite being
a floating point number after division (in Pyton 3). This commit simply coerces `ysize` (and `xsize`) to an int momentarily for use in `np.arange` and `np.ones`.

<hr>

While this commit provides a very simple fix, I'm inclined to apply the integer coercion earlier. Unless I'm mistaken, there is no reason for `ysize` (or `xsize`) to a float. E.g.
```python
xsize = int(self.arrayinfo['xsize'])
ysize = xsize // 2
```

I would even consider adding `from __future__ import division` and replacing all the divisions in the module accordingly (I assume healpy does not need to support anything below Python version 2.6). But that may be felt a bit too much for an initial (simple fix) PR. I put it up here anyway, in case people are actually happy for such a change to be made.

### DIFF
--- a/healpy/projector.py
+++ b/healpy/projector.py
@@ -320,7 +320,8 @@ class GnomonicProj(SphericalProj):
         if self.arrayinfo is None:
             raise TypeError("No projection plane array information defined for "
                             "this projector")
-        xsize,ysize = self.arrayinfo['xsize'],self.arrayinfo['ysize']
+        xsize = int(self.arrayinfo['xsize'])
+        ysize = int(self.arrayinfo['ysize'])
         reso = self.arrayinfo['reso']
         if y is None: x,y = x
         dx = reso/60. * dtor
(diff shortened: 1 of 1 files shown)


zonca:
thanks @evertrol , I think your idea of force `xsize` and `ysize` to int earlier is a good one.

I don't understand your comment on `division`, what would be the advantage of that?

evertrol:
Updated PR

Note: I've used `int(...)` everywhere, but I noticed an `np.long(...)` conversion somewhere in the code. *Except* in the case of Python 2 on a 32 bit platform, these are the same. For this Python 2/32-bit combination, the conversion to `int` leads to a bug if `xsize` or `ysize` is larger than `2**31-1`. I've assumed this never to be the case, but otherwise, this would now be a documented bug. (I'm not sure if such large sizes are sensible anyway, memory-wise or in another way, on a 32 bit platform.)

evertrol:
The comment about division was triggered when I used `//`, and thought about clearing up the C-style (`.../2.0` or `float(x)/float(y)`) floating point division, as Python's current and future default is fp-division for `/`. But that's more an overall stylistic change than an actual fix, so I'll leave it.

zonca:
Thanks this is a great contribution!

PR #368, opened 2016-12-07
Fix warning from floating point division of two integers

keskitalo:
In python3, division operator, "/", produces floating point results even with integer inputs. This produces a warning when the result is used in place of an integer input further on. 

The pull request imports the python3 division and uses it for integer-valued division in fitsfunc.py.

Having the harmless warning removed helps debugging software that uses healpy. This is especially the case when a developer has filtered all warnings to issue exceptions with
```
import warnings
warnings.filterwarnings('error')
```


zonca:
Thanks!

Issue #363, opened 2016-11-04
Python3 map2alm ring weights string conversion error


cjcopi:
map2alm fails to read the ring weights due to a string conversion error.  This is presumably the "usual" Python3 string versus bytes issue, though I have not looked into the details.  map2alm works as expected when run with Python2.

Explicitly, when using
`map2alm(mp, use_weights=True)`
I get (here shown run under a recent anaconda install, I get the same result using a Python 3.4.2 install)
```
TypeError                                 Traceback (most recent call last)
<ipython-input-8-a85dcb21f7a3> in <module>()
----> 1 hp.map2alm(mp,use_weights=True)

/home/copi/tmp/anaconda3/lib/python3.5/site-packages/healpy/sphtfunc.py in map2alm(maps, lmax, mmax, iter, pol, use_weights, datapath)
    155         alms = _sphtools.map2alm(maps, niter = iter, 
    156                                  datapath = datapath, use_weights = use_weights,
--> 157                                  lmax = lmax, mmax = mmax)
    158     else:
    159         # info >= 2 and pol is False : spin 0 spht for each map

_sphtools.pyx in healpy._sphtools.map2alm (healpy/src/_sphtools.cpp:4308)()

TypeError: expected bytes, str found
```

If instead the datapath is explicitly forced to be bytes as in
`map2alm(mp, use_weights=True, datapath=b'/path/to/ring/file/dir')`
I get the error
```
/home/copi/tmp/anaconda3/lib/python3.5/site-packages/healpy/sphtfunc.py in map2alm(maps, lmax, mmax, iter, pol, use_weights, datapath)
    155         alms = _sphtools.map2alm(maps, niter = iter, 
    156                                  datapath = datapath, use_weights = use_weights,
--> 157                                  lmax = lmax, mmax = mmax)
    158     else:
    159         # info >= 2 and pol is False : spin 0 spht for each map

_sphtools.pyx in healpy._sphtools.map2alm (healpy/src/_sphtools.cpp:4357)()

/home/copi/tmp/anaconda3/lib/python3.5/posixpath.py in join(a, *p)
     87                 path += sep + b
     88     except (TypeError, AttributeError, BytesWarning):
---> 89         genericpath._check_arg_types('join', a, *p)
     90         raise
     91     return path

/home/copi/tmp/anaconda3/lib/python3.5/genericpath.py in _check_arg_types(funcname, *args)
    143                             (funcname, s.__class__.__name__)) from None
[...]

PR #369, opened 2016-12-19
Fix for strings in Python 2/3

cjcopi:
This is a proof of concept/description of issue #363 regarding use_weights in map2alm.  I do not expect it is a complete solution.  It has not been tested with Python 2 (I do not have such a test environment) so it may break Python 2.  There are also other cases/assumptions which may be missing.  I do not know much about cython in general or strings+cython at all!

The general idea is that the Python string (datapath variable) is in one format and the C string (c_datapath) needs to be in bytes.  In Python 3 these are different formats and seem to not be automagically converted.

If datapath is a "normal" Python string (unicode and/or type str), then it can be encoded into bytes.  This is the first part of the patch.

If datapath is already in bytes then it is good for the C code, but not for the rest of the Python code, in particular the os.path.join will fail with bytes.  This can be corrected too and is done in the elif block.  This may not be necessary.  It would only come up if the datapath were stored in bytes or if the user provided a path as bytes, instead of a Python string.  I included this because it came up while I was trying to debug the issue.

Finally, I don't know what other possibilities my occur and/or if the user should be protected from not just using a Python string in the first place.

The simplest fix (possibly with a wrapper for Python 2)  is to just use the datapath.encode('UTF-8').  I do think you need to go through a temporary since directly assigning this to c_datapath does not work.  If this were done, then non-Python string input would just fail.

### DIFF
--- a/healpy/src/_sphtools.pyx
+++ b/healpy/src/_sphtools.pyx
@@ -276,13 +276,11 @@ def map2alm(m, lmax = None, mmax = None, niter = 3, use_weights = False,
         if datapath is None:
             datapath = get_datapath()
         # For Python3: datapath must be a str, bdatapath must be bytes
-        if type(datapath) is unicode :
-          bdatapath = datapath.encode('UTF-8')
-        elif isinstance(datapath,bytes) :
-          bdatapath = datapath
-          datapath = datapath.decode('UTF-8')
+        if isinstance(datapath, unicode) :
+            bdatapath = datapath.encode('UTF-8')
         else :
-          print("Now What?")
+            bdatapath = datapath
+            datapath = datapath.decode('UTF-8')
         c_datapath = bdatapath
         weightfile = 'weight_ring_n%05d.fits' % (nside)
         if not os.path.isfile(os.path.join(datapath, weightfile)):
--- a/healpy/src/_sphtools.pyx
+++ b/healpy/src/_sphtools.pyx
@@ -11,7 +11,6 @@ from healpy.pixelfunc import maptype
 import os
 import cython
 from libcpp cimport bool as cbool
-import six
 
 from _common cimport tsize, arr, xcomplex, Healpix_Ordering_Scheme, RING, NEST, Healpix_Map, Alm, ndarray2map, ndarray2alm
 
@@ -276,9 +275,15 @@ def map2alm(m, lmax = None, mmax = None, niter = 3, use_weights = False,
     if use_weights:
         if datapath is None:
             datapath = get_datapath()
-        # b_datapath must be bytes
-        b_datapath = six.b(datapath)
-        c_datapath = b_datapath
+        # For Python3: datapath must be a str, bdatapath must be bytes
+        if type(datapath) is unicode :
+          bdatapath = datapath.encode('UTF-8')
+        elif isinstance(datapath,bytes) :
+          bdatapath = datapath
+          datapath = datapath.decode('UTF-8')
+        else :
+          print("Now What?")
+        c_datapath = bdatapath
         weightfile = 'weight_ring_n%05d.fits' % (nside)
         if not os.path.isfile(os.path.join(datapath, weightfile)):
             raise IOError('Weight file not found in %s' % (datapath))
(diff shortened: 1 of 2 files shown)


lpsinger:
This should use `six.text_type` and `six.binary_type`.

zonca:
thanks @cjcopi , would it be possible for you to use `six.text_type` instead of what you are using? @lpsinger can you show an example?

cjcopi:
I simplified it considerably and am using six.b() to convert to bytes now.  This now does not handle passing in datapath as bytes but I don't know if that is a concern.  Overall, I am unfamiliar with best practices in this area.

zonca:
@lpsinger would you mind to review this? I am not familiar with `six`

lpsinger:
I'm sorry, @cjcopi, I believe that your original solution was correct. I'll post a few comments on it and send you a unit test to add.

healpy/src/_sphtools.pyx
```diff
@@ -275,7 +275,15 @@ def map2alm(m, lmax = None, mmax = None, niter = 3, use_weights = False,
     if use_weights:
         if datapath is None:
             datapath = get_datapath()
-        c_datapath = datapath
+        # For Python3: datapath must be a str, bdatapath must be bytes
+        if type(datapath) is unicode :
+          bdatapath = datapath.encode('UTF-8')
```
lpsinger: Please match the style of the surrounding code and use four spaces for indentation.

healpy/src/_sphtools.pyx
```diff
@@ -275,7 +275,15 @@ def map2alm(m, lmax = None, mmax = None, niter = 3, use_weights = False,
     if use_weights:
         if datapath is None:
             datapath = get_datapath()
-        c_datapath = datapath
+        # For Python3: datapath must be a str, bdatapath must be bytes
+        if type(datapath) is unicode :
```
lpsinger: change to `if isinstance(datapath, unicode):`

healpy/src/_sphtools.pyx
```diff
@@ -275,7 +275,15 @@ def map2alm(m, lmax = None, mmax = None, niter = 3, use_weights = False,
...
+          bdatapath = datapath.encode('UTF-8')
+        elif isinstance(datapath,bytes) :
+          bdatapath = datapath
+          datapath = datapath.decode('UTF-8')
+        else :
```
lpsinger: Get rid of this clause

healpy/src/_sphtools.pyx
```diff
@@ -275,7 +275,15 @@ def map2alm(m, lmax = None, mmax = None, niter = 3, use_weights = False,
     if use_weights:
         if datapath is None:
             datapath = get_datapath()
-        c_datapath = datapath
+        # For Python3: datapath must be a str, bdatapath must be bytes
+        if type(datapath) is unicode :
+          bdatapath = datapath.encode('UTF-8')
+        elif isinstance(datapath,bytes) :
```
lpsinger: Change this to just `else:`

zonca:
ok, thanks, please tag me again once this is ready to merge

lpsinger:
@cjcopi, are you able to complete these changes? Or should I take it from here?

cjcopi:
I have merged and made the suggested changes from @lpsinger.  Hopefully all is well.  With the new semester about to start, if there are other small changes to make it would be quicker for you to handle them yourself rather than waiting on me!

lpsinger:
Thanks! Assuming tests pass, this looks good to go.

lpsinger:
Looks good to me. @zonca, please merge.

lpsinger:
@dpole, that may be a question for @zonca. I did not change the tolerance in my commit.

dpole:
Right! Thanks. Looking a bit more carefully, the assertion in `test_map2alm` and `test_map2alm_pol` are the only two using `atol` instead of `rtol`. FYI, I've tried to use `rtol` with `iter=100`, both tests fails.
Sorry for being a bit off-topic here. Probably I'd better move this discussion in a dedicated issue.

lpsinger:
Yes, please open a new issue.

PR #362, opened 2016-10-29
Add 'lonlat' option to functions in 'pixelfunc'

kadrlica:
This pull requests addresses #361 by adding an optional `lonlat` argument to allow `theta` and `phi` to be provided as longitude and latitude (in degrees) rather than co-latitude and longitude (in radians). The `lonlat` argument was added for `ang2pix`, `pix2ang`, `ang2vec`, `vec2ang`, `get_interp_val`, `get_interp_weights`, and `get_all_neighbours`. Documentation and unit tests were added, along with the utility functions `lonlat2thetaphi` and `thetaphi2lonlat`.

PR #370, opened 2016-12-22
Fix broken doctest introduced by #362

lpsinger:


lpsinger:
Thanks for merging this! Would you mind doing another point release so that the [Debian CI unit tests pass](https://ci.debian.net/packages/h/healpy/)?

zonca:
ok, pushed 1.10.3

PR #377, opened 2017-01-14
Remove NSIDE power of 2 restriction for RING

### LINKED ISSUE #333: Support nside that's not a power of two?
Why does healpy only support nside values that are a power of two?

Is it because the use case of other nside values hasn't come up?
Would pull requests removing the `isnsideok` check from functions where it's not needed be welcome?

This question came up when @tburnett tried to move some of his code over to use `healpy` and tried to call `query_disc` with `nside=12`.

I see that [isnsideok](http://healpy.readthedocs.org/en/latest/generated/healpy.pixelfunc.isnsideok.html) is called for many (all?) healpy functions, and checks that `nside` is a power of two (it's docstring should say that!).

Even for something like [nside2npix](http://healpy.readthedocs.org/en/latest/generated/healpy.pixelfunc.nside2npix.html), where it would be trivial to support `nside` that's not a power of 2.

cc @tburnett

zonca:
Now `check_nside` and `isnsideok` accept a `nest` parameter,
default `False`, it checks that `nside` is a power of 2 only
if `nest=True`.

Closes #333 

@lpsinger @mreineck (@cdeil?) can you take a look at it?

### DIFF
--- a/healpy/pixelfunc.py
+++ b/healpy/pixelfunc.py
@@ -941,8 +941,11 @@ def nside2order(nside):
     True
 
     >>> hp.nside2order(7)
-    2
+    Traceback (most recent call last):
+        ...
+    ValueError: 7 is not a valid nside parameter (must be a power of 2, less than 2**30)
     """
+    check_nside(nside, nest=True)
     return len('{0:b}'.format(nside)) - 1
 
 def nside2resol(nside, arcmin=False):
@@ -978,7 +981,6 @@ def nside2resol(nside, arcmin=False):
 
     >>> hp.nside2resol(7)
     0.1461895297066412
-
     """
     
     resol = np.sqrt(nside2pixarea(nside))
--- a/healpy/pixelfunc.py
+++ b/healpy/pixelfunc.py
@@ -1018,6 +1018,9 @@ def nside2pixarea(nside, degrees=False):
 
     >>> hp.nside2pixarea(256)
     1.5978966540475428e-05
+
+    >>> hp.nside2pixarea(7)
+    0.021371378595848933
     """
     
     pixarea = 4*np.pi/nside2npix(nside)
--- a/healpy/pixelfunc.py
+++ b/healpy/pixelfunc.py
@@ -1061,7 +1061,7 @@ def npix2nside(npix):
         raise ValueError("Wrong pixel number (it is not 12*nside**2)")
     return int(np.sqrt(npix/12.))
 
-def order2nside(order, nest=False):
+def order2nside(order):
     """Give the nside parameter for the given resolution order.
 
     Parameters
@@ -1093,7 +1093,7 @@ def order2nside(order, nest=False):
     ValueError: 2147483648 is not a valid nside parameter (must be a power of 2, less than 2**30)
     """
     nside = 1<<order
-    check_nside(nside, nside=nside)
+    check_nside(nside, nest=True)
     return nside
 
 def isnsideok(nside, nest=False):
--- a/healpy/pixelfunc.py
+++ b/healpy/pixelfunc.py
@@ -425,7 +425,7 @@ def ang2pix(nside,theta,phi,nest=False,lonlat=False):
     if lonlat:
         theta,phi = lonlat2thetaphi(theta,phi)
     check_theta_valid(theta)
-    check_nside(nside)
+    check_nside(nside, nest=nest)
     if nest:
         return pixlib._ang2pix_nest(nside,theta,phi)
     else:
@@ -471,7 +471,7 @@ def pix2ang(nside,ipix,nest=False,lonlat=False):
     >>> hp.pix2ang([1, 2, 4, 8], 11, lonlat=True)
     (array([ 315. ,  337.5,  337.5,  337.5]), array([-41.8103149 ,  41.8103149 ,  66.44353569,  78.28414761]))
     """
-    check_nside(nside)
+    check_nside(nside, nest=nest)
     if nest:
         theta,phi = pixlib._pix2ang_nest(nside, ipix)
     else:
@@ -515,7 +515,7 @@ def xyf2pix(nside,x,y,face,nest=False):
     >>> hp.xyf2pix(16, [8, 8, 8, 15, 0], [8, 8, 7, 15, 0], [4, 0, 5, 0, 8])
     array([1440,  427, 1520,    0, 3068])
     """
-    check_nside(nside)
+    check_nside(nside, nest=nest)
     if nest:
         return pixlib._xyf2pix_nest(nside,x,y,face)
     else:
@@ -556,7 +556,7 @@ def pix2xyf(nside,ipix,nest=False):
     >>> hp.pix2xyf([1, 2, 4, 8], 11)
     (array([0, 1, 3, 7]), array([0, 0, 2, 6]), array([11,  3,  3,  3]))
     """
-    check_nside(nside)
+    check_nside(nside, nest=nest)
     if nest:
         return pixlib._pix2xyf_nest(nside, ipix)
     else:
@@ -635,7 +635,7 @@ def pix2vec(nside,ipix,nest=False):
     >>> hp.pix2vec([1, 2], 11)
     (array([ 0.52704628,  0.68861915]), array([-0.52704628, -0.28523539]), array([-0.66666667,  0.66666667]))
     """
-    check_nside(nside)
+    check_nside(nside, nest=nest)
     if nest:
         return pixlib._pix2vec_nest(nside,ipix)
     else:
@@ -733,7 +733,7 @@ def ring2nest(nside, ipix):
     >>> hp.ring2nest([1, 2, 4, 8], 11)
     array([ 11,  13,  61, 253])
     """
-    check_nside(nside)
+    check_nside(nside, nest=False)
     return pixlib._ring2nest(nside, ipix)
 
 def nest2ring(nside, ipix):
@@ -767,7 +767,7 @@ def nest2ring(nside, ipix):
     >>> hp.nest2ring([1, 2, 4, 8], 11)
     array([ 11,   2,  12, 211])
     """
-    check_nside(nside)
+    check_nside(nside, nest=True)
     return pixlib._nest2ring(nside, ipix)
 
 @accept_ma
@@ -906,13 +906,7 @@ def nside2npix(nside):
 
     >>> np.all([hp.nside2npix(nside) == 12 * nside**2 for nside in [2**n for n in range(12)]])
     True
-
-    >>> hp.nside2npix(7)
-    Traceback (most recent call last):
-        ...
-    ValueError: 7 is not a valid nside parameter (must be a power of 2, less than 2**30)
     """
-    check_nside(nside)
     return 12*nside**2
 
 def nside2order(nside):
@@ -942,13 +936,7 @@ def nside2order(nside):
 
     >>> np.all([hp.nside2order(2**o) == o for o in range(30)])
     True
-
-    >>> hp.nside2order(7)
-    Traceback (most recent call last):
-        ...
-    ValueError: 7 is not a valid nside parameter (must be a power of 2, less than 2**30)
     """
-    check_nside(nside)
     return len('{0:b}'.format(nside)) - 1
 
 def nside2resol(nside, arcmin=False):
@@ -982,12 +970,7 @@ def nside2resol(nside, arcmin=False):
     >>> hp.nside2resol(256)
     0.0039973699529159707
 
-    >>> hp.nside2resol(7)
-    Traceback (most recent call last):
-       ...
-    ValueError: 7 is not a valid nside parameter (must be a power of 2, less than 2**30)
     """
-    check_nside(nside)
     
     resol = np.sqrt(nside2pixarea(nside))
 
@@ -1024,13 +1007,7 @@ def nside2pixarea(nside, degrees=False):
 
     >>> hp.nside2pixarea(256)
     1.5978966540475428e-05
-
-    >>> hp.nside2pixarea(7)
-    Traceback (most recent call last):
-       ...
-    ValueError: 7 is not a valid nside parameter (must be a power of 2, less than 2**30)
     """
-    check_nside(nside)
     
     pixarea = 4*np.pi/nside2npix(nside)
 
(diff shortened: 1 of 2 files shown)


healpy/pixelfunc.py
```diff
@@ -733,7 +733,7 @@ def ring2nest(nside, ipix):
     >>> hp.ring2nest([1, 2, 4, 8], 11)
     array([ 11,  13,  61, 253])
     """
-    check_nside(nside)
+    check_nside(nside, nest=False)
```
mreineck: This should be nest=True, I think, otherwise you cannot convert to nested scheme.

healpy/src/_pixelfunc.pyx
```diff
@@ -100,10 +100,5 @@ def pix2ring(nside, np.ndarray[int64, ndim=1] pix not None, nest=False):
...
 
-cdef bool isnsideok(int nside):
-    if nside < 0 or nside != 2**int(round(np.log2(nside))):
-        return False
-    else:
-        return True
-
-
+cdef bool isnsideok(int nside, bool nest=False):
+    return (nest is False) or (nside > 0 and nside == 2**int(round(np.log2(nside))))
```
mreineck: nside must be >0 even when nest is False.

healpy/src/_pixelfunc.pyx
```diff
@@ -41,7 +41,7 @@ def ringinfo(nside, np.ndarray[int64, ndim=1] ring not None):
            -0.33333333, -0.66666667]), array([ 0.        ,  0.39965263,  0.74535599,  0.94280904,  1.        ,
             0.94280904,  0.74535599]), array([ True,  True,  True, False,  True, False,  True], dtype=bool))
     """
-    if not isnsideok(nside):
+    if not isnsideok(nside, nest=True):
```
mreineck: This whole function should be done on RING-ordered Healpix bases, so that it works for all Nsides.

lpsinger:
Do any of the pix2ang or ang2pix functions work for nside being not a power of 2?

mreineck:
> Do any of the pix2ang or ang2pix functions work for nside being not a power of 2?

The C++ ones certainly do, and as far as I see they are being called by healpy.

lpsinger:
> The C++ ones certainly do, and as far as I see they are being called by healpy.

Then the `pix2ang`, `ang2pix`, and all related functions should pass the `nest` argument through to `check_nside` and we need some unit tests for non-powers-of-2.

mreineck:
Absolutely. But isn't this already done in the pull request?

mreineck:
To be honest, I'd be much in favor of removing the default on the "nest" key word in "check_nside" altogether; then we would immediately notice all callers where it is missing.
Same for many, many other defaults all around the package. But probably that's just my C++ background ...

zonca:
@mreineck that would be too cumbersome for interactive usage.

@lpsinger there are some doctests in `isnsideok`

zonca:
@lpsinger would you like to review this PR as well?

mreineck:
> @mreineck that would be too cumbersome for interactive usage.

I.e. in the usage scenario where there is the highest probability of inadvertent mistakes?
Maybe I'm not the typical user, but I'd really prefer an immediate and clear error message and a few more key strokes to non-obvious defaults, which cause me to silently compute things that are different from what I expect.
And choosing one of RING or NESTED over the other for a default is definitely non-obvious in my book. In some applications (e.g. everything to do with spherical harmonics) RING is the clear favorite, whereas in VO database-like applications no one would use anything except NESTED. Whatever default you pick, one community will be confused.

But again, I'm not really involved in the Python part of healpy and just wanted to give my impressions. I'll shut up now :)

zonca:
@mreineck my preference would be to have a single point where you can switch between RING or NESTED. Generally in a single script you either always use ring or always use nest, passing to all functions `nest=True` or `nest=False` is tedious.
For convenience I always use RING when I use `healpy`, even if all our maps are NESTED and I use NESTED for pointing, I just convert to RING at the beginning and end.

Brainstorming down here:

* I'd like to have something like `hp.set_ordering(nest)` and that influences all the subsequent functions, maybe trigger a warning if you switch ordering in the middle of a script or if you call a function without having set this beforehand.
* or access all pixel based functions through a object so you can instantiate one `nest_nside32_pixel_manager = hp.PixelManager(nest=True, nside=32)`, `pix = nest_nside32_pixel_manager.ang2pix(theta, phi)`

cdeil:
First of all - @zonca - thanks for adding this!!!

I think most `healpy` users read and write code using `healpy` every few weeks or months or years,  i.e. infrequently, and it's usually just a few function calls. So I think the "it's annoying to have to repeat the `nest` option many times" is only an issue for few power-users that read and write a lot of healpy code.

I'd be +1 to follow @mreineck's suggestion and remove the default altogether.

@zonca - concerning your suggestions to add a convenience for users to choose the default without having to repeat themselves:
- `hp.set_ordering(nest)` - that seems error-prone, I think it's a bad idea.
- `hp.PixelManager(nest=True, nside=32)` - that sounds similar to what @ziotom78 did in https://github.com/ziotom78/Healpix.jl#dealing-with-resolutions . @ziotom78 - what's your experience with it? Thoughts on this API question for healpy?
- Overall it's a tradeoff between convenience, and creating a larger API, i.e. two ways to handle that setting. Like I said, my suggestion would be to err on the side of caution and *not* add a default or second way to handle this, and just explain in the docs something like:
```
# If you always want the same `nest`, then define it once at the top
nest = True
# and then use it everywhere
hp.something(..., nest=nest)
np.something_else(..., nest=nest)
```

I haven't looked at the code changes yet. Does this PR or any of the solutions discussed here introduce regressions, i.e. can break existing scripts? If so, I think it's a non-starter, right?

healpy/pixelfunc.py
```diff
@@ -1110,12 +1084,14 @@ def order2nside(order):
     ValueError: 2147483648 is not a valid nside parameter (must be a power of 2, less than 2**30)
     """
     nside = 1<<order
-    check_nside(nside)
+    check_nside(nside, nside=nside)
```
cdeil: Here you write
```
check_nside(nside, nside=nside)
```
This is a bug, no?

mreineck:
This is slowly getting off-topic for this pull request, but here's my take on the "how should we specify nest/ring without having to type too much?" issue...
I think the current dilemma stems from the fact that the healpy interface is "function-oriented" rather than object-oriented. If you want a specific job done, you call the appropriate function, and you supply information like nside, ordering, etc. over and over again. In contrast, Healpix C++ solves this by introducing the Healpix_Base class, which is constructed with a given Nside and ordering and then remembers these parameters.
So using this approach, you would first create a Healpix_Base object with the nside and ordering that suits your needs, and from then on just call its methods, which do the right thing without requiring specification of ordering and nside all the time.
(You can have as many different Healpix_Base instances at a time as you like.)
Of course, that would be a big departure from the current interface presented by healpy, but I seriously think it could be worth it.
BTW, if there is interest, we should probably move this discussion to a new issue!

ziotom78:
@zonca, @cdeil: In [Healpix.jl](https://github.com/ziotom78/Healpix.jl), I require the user to build a [`Resolution`](https://github.com/ziotom78/Healpix.jl/blob/master/src/Healpix.jl#L263) object (which has no knowledge of the RING/NEST dichotomy) in order to do transformations like angles to pixels, etc.. For each transformation, the package implements three functions; taking as example `healpy.ang2pix`, Healpix.jl implements the following functions:

1. Functions [`ang2pixNest`](https://github.com/ziotom78/Healpix.jl/blob/master/src/Healpix.jl#L397) and [`ang2pixRing`](https://github.com/ziotom78/Healpix.jl/blob/master/src/Healpix.jl#L526); these need a `Resolution` object as their first parameter;
2. Function [`ang2pix`](https://github.com/ziotom78/Healpix.jl/blob/master/src/Healpix.jl#L891) requires a `Map` object, which embeds the ordering information within the type itself.

This meets a few requirements I had in mind when I designed Healpix.jl's API:

1. When I am working with a Healpix map, I do not want to explicitly deal with the RING/NEST stuff and just use whatever ordering my map uses. (This includes unwanted conversions from NEST to RING during file loads, like [`healpy.read_map`](https://github.com/healpy/healpy/blob/master/healpy/fitsfunc.py#L359) does when using the default `nest=False`.) Healpix.jl's achieves this goal because (1) [`readMapFromFITS`](https://github.com/ziotom78/Healpix.jl/blob/master/src/Healpix.jl#L725) returns different objects, depending on the ordering of the file being loaded; (2) Julia's [multiple dispatch](https://en.wikipedia.org/wiki/Multiple_dispatch) ensures that functions accepting objects of type `Map` receive the ordering as part of the function's signature (e.g., see [`ang2pix`](https://github.com/ziotom78/Healpix.jl/blob/master/src/Healpix.jl#L891), mentioned above). See the definition of type [`Order`](https://github.com/ziotom78/Healpix.jl/blob/master/src/Healpix.jl#L673).

2. When I am just working with directions and points on the celestial sphere but have no map loaded, I just create a `Resolution` object, which is conceptually equivalent to @zonca's `PixelManager`. I usually stick to one ordering which is decided since the very beginning: therefore, there is no significant cognitive burden in calling `ang2pixNest`/`ang2pixRing` instead of `ang2pix`.

This has the disadvantage that the object returned by `readMapFromFITS` is not a plain array, but a object of type [`Map`](https://github.com/ziotom78/Healpix.jl/blob/master/src/Healpix.jl#L692); thus, in order to access the value of the pixel, one has to explicitly mention the `pixels` field:

`````julia
import Healpix

m = Healpix.readMapFromFITS('wmap_512.fits', 1, Float64)
print(mean(m.pixels))
`````
 
Moreover, keep in mind that Healpix.jl has a very little userbase: I have not received any feedback about this way of dealing with resolutions and orderings, yet.

From a broader point of view, I think that Healpix.jl's approach is better suited for a language like Julia, which determines which function to call depending on the type of the parameters. I do not think my approach should be used in Healpy, however: it would cause backward-incompatible changes and confuse long-time users. Personally, I have always thought of Healpy's approach to the ordering problem (stick with one ordering – RING – by default) interesting and well suited for working in the REPL; I would not change it.

**Comment by lpsinger** [comment]
> This is slowly getting off-topic for this pull request, but here's my take on the "how should we specify nest/ring without having to type too much?" issue...
> I think the current dilemma stems from the fact that the healpy interface is "function-oriented" rather than object-oriented. If you want a specific job done, you call the appropriate function, and you supply information like nside, ordering, etc. over and over again. In contrast, Healpix C++ solves this by introducing the Healpix_Base class, which is constructed with a given Nside and ordering and then remembers these parameters.
> So using this approach, you would first create a Healpix_Base object with the nside and ordering that suits your needs, and from then on just call its methods, which do the right thing without requiring specification of ordering and nside all the time.
> (You can have as many different Healpix_Base instances at a time as you like.)
> Of course, that would be a big departure from the current interface presented by healpy, but I seriously think it could be worth it.
> BTW, if there is interest, we should probably move this discussion to a new issue!

I have a patch for Astropy by that implements different HEALPix indexing schemes as `BaseRepresentation` subclasses so that any `SkyCoord` can be converted to/from HEALPix. I could supply this code as a Healpy pull request.

File: healpy/pixelfunc.py [hunk 5]
```diff
@@ -906,13 +906,7 @@ def nside2npix(nside):
 
     >>> np.all([hp.nside2npix(nside) == 12 * nside**2 for nside in [2**n for n in range(12)]])
     True
-
-    >>> hp.nside2npix(7)
```
lpsinger: Instead of removing this check, could we check that it returns the correct answer, 12*7**2 = 588?

healpy/pixelfunc.py
```diff
@@ -942,13 +936,7 @@ def nside2order(nside):
 
     >>> np.all([hp.nside2order(2**o) == o for o in range(30)])
     True
-
-    >>> hp.nside2order(7)
-    Traceback (most recent call last):
```
lpsinger: This should still raise an exception because the concept of an order is tied to nside being a power of 2. Please bring back this doctest.

healpy/pixelfunc.py
```diff
@@ -982,12 +970,7 @@ def nside2resol(nside, arcmin=False):
     >>> hp.nside2resol(256)
     0.0039973699529159707
 
-    >>> hp.nside2resol(7)
-    Traceback (most recent call last):
```
lpsinger: Also, let's keep this check but assert that it gives the correct answer instead of raising an exception.

healpy/pixelfunc.py
```diff
@@ -1024,13 +1007,7 @@ def nside2pixarea(nside, degrees=False):
 
     >>> hp.nside2pixarea(256)
     1.5978966540475428e-05
-
-    >>> hp.nside2pixarea(7)
-    Traceback (most recent call last):
```
lpsinger: Also bring this doctest back.

healpy/pixelfunc.py
```diff
@@ -1071,14 +1048,11 @@ def npix2nside(npix):
...
-    if nside != np.floor(nside):
+    if not isnpixok(npix):
         raise ValueError("Wrong pixel number (it is not 12*nside**2)")
-    nside=int(np.floor(nside))
-    check_nside(nside)
-    return nside
+    return int(np.sqrt(npix/12.))
 
-def order2nside(order):
+def order2nside(order, nest=False):
```
lpsinger: order2nside should not take a `nest` argument.

healpy/pixelfunc.py
```diff
@@ -944,11 +941,8 @@ def nside2order(nside):
     True
 
     >>> hp.nside2order(7)
-    Traceback (most recent call last):
```
lpsinger: This *should* raise a ValueError.

healpy/pixelfunc.py
```diff
@@ -983,11 +977,9 @@ def nside2resol(nside, arcmin=False):
     0.0039973699529159707
 
     >>> hp.nside2resol(7)
-    Traceback (most recent call last):
-       ...
-    ValueError: 7 is not a valid nside parameter (must be a power of 2, less than 2**30)
+    0.1461895297066412
+
```
lpsinger: One too many newlines here?

healpy/pixelfunc.py
lpsinger: @zonca, looks like you forgot to restore this doctest?

PR #382, opened 2017-02-24
Remove unused code and tweak _pixelfunc.pyx a little bit

mreineck:
This removed the unused interface code to the Healpix FITS functionality.
I also found an inefficient implementation of "isnsideok" in _pixelfunc.pyx and synchronized it with the implementation used elsewhere in the code.


healpy/src/_pixelfunc.pyx
```diff
@@ -88,11 +88,7 @@ def pix2ring(nside, np.ndarray[int64, ndim=1] pix not None, nest=False):
     if not isnsideok(nside, nest=nest):
         raise ValueError('Wrong nside value, must be a power of 2 (for nested), less than 2**30')
     cdef Healpix_Ordering_Scheme scheme
```
zonca: @mreineck can we also remove `scheme` from here?

mreineck:
Yes, that was an oversight. Should be fixed now!

PR #386, opened 2017-04-13
Add overwrite keyword to write_map

keskitalo:
Changes to fitsfunc.py in 443855896acfbbc1602a96e55126b022c677a133 caused a change in write_map() behavior. Prior, trying to write to a file that already existed caused the file to be silently overwritten.

After the change, write_map throws an IOError when the user attempts to write to an existing file. The only way to avoid the error is to check for the file existence and remove it before write_map().

It seems unnecessary to require users to check file existence every time they write a map. This pull request implements a keyword argument to write_map() called "overwrite". The current default behavior (fail when file exists) can be retained by setting overwrite=False by default. Defaulting it to True would restore the old default behavior and not require packages that depend on Healpy to be changed everywhere write_map() is called. Either way, user can choose to overwrite existing files without explicitly checking their existence.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,7 @@
+Next Release Candidate
+
+* Change in write_map default behavior: <https://github.com/healpy/healpy/pull/379> and <https://github.com/healpy/healpy/pull/386>
+
 Release 1.10.1, 8 Nov 2016
 
 * Removed support for Python 2.6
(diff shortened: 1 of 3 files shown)


lpsinger:
This matches the behavior of Astropy's [`HDUList.writeto`](http://docs.astropy.org/en/v1.3.2/io/fits/api/hdulists.html#astropy.io.fits.HDUList.writeto) and [`Table.writeto`](http://docs.astropy.org/en/v1.3.2/io/unified.html#writing) methods.

But due to the change in the default behavior, you should document this in the changelog.

keskitalo:
I added the current release candidate to the changelog and made a note about the change in default behaviour. Thanks for the links to astropy, I didn't realize the type of the error was different for Python 2/3.

keskitalo:
Turns out the January modification caused a healpy write_map unit test to fail when called repeatedly. I modified the test to use the new "overwrite" keyword.

zonca:
thanks!

PR #387, opened 2017-04-21
Fix deprecated comparison that will fail in the future

keskitalo:
I keep getting a warning because of this comparison. None should be checked using "is" rather than "==".

### DIFF
--- a/healpy/pixelfunc.py
+++ b/healpy/pixelfunc.py
@@ -1280,7 +1280,7 @@ def get_interp_weights(nside,theta,phi=None,nest=False,lonlat=False):
            [ 0.25,  0.  ]]))
     """
     check_nside(nside, nest=nest)
-    if phi == None:
+    if phi is None:
         theta,phi = pix2ang(nside,theta,nest=nest)
     elif lonlat:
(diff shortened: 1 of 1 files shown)


cdeil:
This is now giving an error with numpy 1.13.3 and healpy 1.10.3 in Macports when passing a numpy array for `phi`:
```
In [7]: hp.get_interp_weights(1, 0, np.array([0, 0]))
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-7-ff2fafd20666> in <module>()
----> 1 hp.get_interp_weights(1, 0, np.array([0, 0]))

/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/healpy/pixelfunc.py in get_interp_weights(nside, theta, phi, nest, lonlat)
   1293     """
   1294     check_nside(nside)
-> 1295     if phi == None:
   1296         theta,phi = pix2ang(nside,theta,nest=nest)
   1297     elif lonlat:

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
@lpsinger - Could you please update to healpy 1.11 in Macports?

Issue #322, opened 2016-01-26
cartview lonra bounds bug


gregreen:
In visufunc.cartview, I sometimes get the following error:

```
TypeError: Wrong argument lonra or latra. Must be lonra=[a,b],latra=[c,d] with a<b, c<d, a>=-180, b<=180, c>=-90, d<=+90
```

The restrictions on `lonra` are incorrect: `a<b` should not be required (for that matter, requiring `a>=-180` and `b<=180` is also incorrect), because longitude obeys modulo-360 arithmetic.

This makes it difficult to plot sections of the sky that span longitude = 180 degrees without resorting to fiddling with the `rot` parameter.

zonca:
sorry, I don't have much experience with the projections. do you think that changing or removing the checks in the code would be enough? https://github.com/healpy/healpy/blob/029bb260239621f19d2fdcf036d4c4f5330b3af5/healpy/projector.py#L556

zonca:
@gregreen if you are interested in following up, please reopen the issue

gregreen:
Sorry for not following up on this. Dealing with longitudes in projections is a bit tricky. The problem is that adding 360 deg to a longitude does not change the meaning. I think the code needs two changes:

1. After [this line](https://github.com/healpy/healpy/blob/029bb260239621f19d2fdcf036d4c4f5330b3af5/healpy/projector.py#L553), put an else block that shifts ``lonra[1]`` into the range ``[lonra[0], lonra[0]+360]``.
2. Remove the check on ``lonra`` in [this line](https://github.com/healpy/healpy/blob/029bb260239621f19d2fdcf036d4c4f5330b3af5/healpy/projector.py#L556), the ``lonra`` checks should be removed altogether. Any combination of two longitudes is a valid specification (even if the second longitude is less than the first longitude - imagine ``[179, 1]`` as a longitude range).

zonca:
thanks @gregreen can you make up a simple snippet for testing this?

PR #388, opened 2017-04-23
Fix longitude bounds in cartview projection

zonca:
See #322 

@gregreen can you please test my implementation?

even better if you have a quick test we can add to the unit tests.


zonca:
@gregreen could you please crosscheck the implementation?

Issue #86, opened 2013-02-20
add testing to rotate_alm


zonca:
we need to add  more testing for `rotate_alm`, there is currently just a unit test about consistency of forward and backward rotation, but we would need a direct comparison with a rotation made in `healpix`

PR #399, opened 2017-07-07
Coordsys2euler zyz

keskitalo:
The rotate_alm method in healpy/src/_sphtools.pyx mentions a helper routine, coordsys2euler_zyz that can be used to obtain the euler angles corresponding to a coordinate system transformation. However, the routine does not exist in Healpy but is rather implemented only in the Healpix Fortran library. This pull request implements the missing routine.

The angles are needed because the Cython interface does not expose the rotation matrix version of rotate_alm in the C++ library. The rotation matrices are a separate class in the C++ library that would need to be exported to use the C++ routines directly. It is much more straightforward to convert the rotation matrices into Euler angles in the Python side.

This pull request also adds the "overwrite" keyword to write_cl (to mirror write_map), generalizes the accepted inputs to write_cl to include 2D arrays and fixes trivial white space issues in the affected files.

The unit tests in coordsys2euler_zyz were checked against the Fortran library.

### DIFF
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -62,7 +62,7 @@ def read_cl(filename, dtype=np.float64, h=False):
     else:
         return cl
 
-def write_cl(filename, cl, dtype=np.float64):
+def write_cl(filename, cl, dtype=np.float64, overwrite=False):
     """Writes Cl into an healpix file, as IDL cl2fits.
 
     Parameters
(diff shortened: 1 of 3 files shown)


keskitalo:
The issue was raised in #86 thread.

keskitalo:
The latest commit addresses #86 and adds a unit test that compares healpy rotate_alm to values from the Fortran library.

zonca:
Awesome, thanks!

Issue #113, opened 2013-08-07
Returning (I,Q,U) tuples makes it really easy to make stupid mistakes that go unnoticed. 


marius311:
Consider the line, 

```
alm = map2alm(map1) + map2alm(map2) 
```

If map1 and map2 are temperature maps only, the return value is an array which is the sum of alms. If map2 and map2 are an (I,Q,U) tuple of maps, the each call returns a tuple, which when added are concatenated into a length-6 tuple, rather than adding the maps. This is a pretty easy mistake to make which is worsened by the fact that if you then call alm2cl(alm) with alm as the length-6 tuple, you don't get an error message (you get the 21 cross spectra). You can make the same mistake with several other function combinations, e.g. synfast/anafast, and operators e.g. +, *, etc..   

A proposed solution would be to make map2alm (and all other such functions) return 2-d arrays, that way the addition operator works the same whether you have just a T map or I,Q,U maps. It would not break anyone's code that was indexing the return values, i.e. map2alm(map)[0] would still work the same.

zonca:
implemented by @keskitalo

PR #400, opened 2017-07-10
Tuple2array

keskitalo:
This pull request addresses #113. All multi-component map and alm return values are changed into 2D numpy arrays instead of tuples. For the user, the change is mostly transparent as the 2D arrays handle like tuples in almost all cases. The tuple concatenation (referenced in #113) is an exception and demonstrates that the 2D array behavior is more intuitive. 

The change makes it unnecessary to cast the maps and alms into arrays in the client side for basic operations like addition and multiplication and should result in cleaner client code.

There is one use of read_map that may break existing codes. When the header is requested (h=True), old read_map would return a tuple of map components and the header appended last. The new code returns a 2-element tuple: (map, header) regardless of the number of map components. There was no unit test for this particular way of calling read_map but one was added.

In addition:
- read_alm is extended to accept multiple HDU:s to read from
- write_alm is extended to accept the "overwrite" keyword to match write_map.
- two instances of pixelfunc.npix2nside(len(map_in[0])) were replaced with the shorthand get_nside(map_in) in smoothing
- added unit test for the the variant of read_map that also returns the header
- modified the reorder doctest to account for the fact that a masked multidimensional map is also a single object rather than a tuple of masked components
- fixed one typo and number of trivial white space issues

### DIFF
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -391,7 +391,8 @@ def read_map(filename,field=0,dtype=np.float64,nest=False,partial=False,hdu=1,h=
         else:
             return ret[0]
     else:
-        ret = np.array(ret)
+        if all(dt == dtype[0] for dt in dtype):
+            ret = np.array(ret)
         if h:
             return ret, header
         else:
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -56,7 +56,7 @@ def read_cl(filename, dtype=np.float64, h=False):
       the cl array
     """
     fits_hdu = _get_hdu(filename, hdu=1)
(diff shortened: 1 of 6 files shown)


lpsinger:
Won't this break if the fields have different dtypes?

keskitalo:
Excellent question! Numpy.array automatically upcasts the provided vectors into a common datatype that can hold all objects.

This is mostly an issue with read_map which, by default, already casts everything to np.float64. If the user does provide a tuple of data types for different fields, read_map should not try to pack everything into a single 2D array. I'll push a fix shortly.

keskitalo:
Did @lpsinger have another example in mind of packing vectors into an array causing unexpected behavior?

zonca:
thanks, this looks good, I'm merging it

PR #402, opened 2017-07-25
Fix mollzoom crash

keskitalo:
This pull request fixes an error that caused mollzoom to throw an exception. 

When calling mollzoom with a partial sky and hist=True, an error was generated if the stamp at the center of the sky contained no valid pixels. Similar issues would have been caused by other projections when the viewable sky contained no valid pixels.

The pull request also adds relevant unit tests and fixes trivial white space issues in the affected files.

### DIFF
--- a/healpy/projaxes.py
+++ b/healpy/projaxes.py
@@ -1,22 +1,22 @@
-# 
+#
 #  This file is part of Healpy.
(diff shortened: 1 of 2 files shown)


zonca:
thanks!

PR #407, opened 2017-08-08
Handle module __builtins__ both as dict or module

### LINKED ISSUE #391: AttributeError about __NUMPY_SETUP__ from setup.py
@adl1995 and I are starting a Python project using `healpy`, and while getting set up we got this error:
```
  File "/var/folders/my/m6ynh3bn6tq06h7xr3js0z7r0000gn/T/easy_install-eq87v75w/healpy-1.10.3/setup.py", line 353, in finalize_options
AttributeError: 'dict' object has no attribute '__NUMPY_SETUP__'
```
Full log: https://travis-ci.org/hipspy/hips/jobs/236887749#L295

Now I think the issue is simply that we're missing `numpy` from the environment where we tried to `pip install healpy` (not sure yet, still looking). But even if that's the case: the error message could be improved, i.e. say "first install numpy" instead of an `AttributeError`?

zonca:
closes #391

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,6 +60,11 @@ else:
     log.info('%s is installed; using Cython')
     from Cython.Distutils import build_ext, Extension
 
+def set_builtin(name, value):
+    if isinstance(__builtins__, dict):
+        __builtins__[name] = value
+    else:
+        setattr(__builtins__, name, value)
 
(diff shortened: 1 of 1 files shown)


lpsinger:
I'm sorry, I only just saw this.

PR #415, opened 2017-10-12
Add nlocs argument to the mollview to control the number of ticks in…

heccphy:
… the colorbar

I did the following modification to visufunc.mollview and projaxes.BaundaryLocator:
    1. Add an argument "nlocs=2" to mollview to set the number of ticks
    2. Add an argument "norm=None" to projaxes.BaundaryLocator so that it can
properly set the ticks in both linear and log scale.

If you find this useful, you may want to improve this feature by adding norm="hist" to projaxes.BaundaryLocator and also fixe other functions in visufunc.py

### DIFF
--- a/healpy/projaxes.py
+++ b/healpy/projaxes.py
@@ -795,17 +795,23 @@ def create_colormap(cmap):
 #   A Locator that gives the bounds of the interval
 #
 class BoundaryLocator(matplotlib.ticker.Locator):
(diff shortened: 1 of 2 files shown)


zonca:
very nice! I tested it here: https://gist.github.com/zonca/b5ef12be4faae62e98f71d71d0808b57

heccphy:
Thank you for letting me know!

PR #419, opened 2017-10-29
Compare to None using is instead of ==

cdeil:
@lpsinger - This is just to follow up to https://github.com/healpy/healpy/pull/387#issuecomment-340307854 , I grepped `healpy` that there are no other cases where you have `==None` except for this one.


zonca:
thanks

PR #440, opened 2018-05-17
Make docstrings reproducible with bleeding-edge versions of Numpy

lpsinger:
Numpy's string representation of arrays changed significantly from 1.13 to 1.14. See the [changelog]. We have to set the legacy representation mode for our doctests to be reproducible across both old and new versions of Numpy.

There is a lingering issue with the doctest for `hp.reorder` that is due to a regression in Numpy. See numpy/numpy#11112.

[changelog]: https://docs.scipy.org/doc/numpy-1.14.0/release.html#many-changes-to-array-printing-disableable-with-the-new-legacy-printing-mode

### DIFF
--- a/healpy/test/conftest.py
+++ b/healpy/test/conftest.py
@@ -1,0 +1,13 @@
+import numpy as np
+import pytest
+
+
+def pytest_configure():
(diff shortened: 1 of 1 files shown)


zonca:
thanks

Issue #356, opened 2016-10-17
Improve accuracy of Healpix map analysis


mreineck:
I have been working for a long time to improve the accuracy of the Healpix map2alm operation. In its current state there are two approaches:
- without any weigts, and
- with ring weights

These both produce only approximate results for all multipole moments.
A few weeks ago I managed to compute new sets of weights that vary from pixel to pixel. If these are used, Healpix-discretized functions on the sphere which have a band limit of lmax=3/2*Nside can be recovered by map2alm with machine precision. If higher multipoles are present, the result is affected by aliasing, but it should still be better than with the previous methods.

Computing the weight sets for all required resolutions is expensive, but they can be stored and read from disk when needed, similar to the approach used with the ring weights. Due to the various symmetries of the Healpix grid, the size of a weight set at a given Nside is about 1/16th of a full Healpix map.

Is there interest in using these new weights from healpy? If yes, it's probably simplest to just write the methods for reading the weights from disk and applying them to a map in Python. I can provide the FITS files if desired.

cjcopi:
I recently "rediscovered" the inaccuracies and  think this would be a great addition.

Is there a reason to not include this in HEALPix and pick up the addition that way?

Regardless, this would be of great value and I would be happy to (have a student) do some testing/start using the new weights.

zonca:
yes, sure, can you post the weights publicly so people can test it?
what is the size of the full set?
If too large we should probably implement a way to download them after installation.

mreineck:
I can certainly provide you with a few weight sets. Unfortunately, the full set is too large to be distributed with the package. Going up to Nside=4096, we are talking about several 100MB.
I have software in the Healpix repository to compute the weight sets on demand and store them to disk, but for high resolutions this requires a very beefy machine (roughly a few CPU minutes at Nside=128, but hundreds of CPU hours at Nside=2048). So I'm not sure what the best distribution strategy for the data would be.

zonca:
I think the best would be to store them publicly, possibly inside another
GitHub repository and then Implement a way to download only one specific
set as needed.

lpsinger:
How long are we talking for generating them from scratch?

mreineck:
As I said, roughly a minute for Nside=128, but the cost scales with something like Nside^4. I can't give hard numbers on the complexity since there is a CG iteration involved, and I can't predict the necessary number of iterations.

lpsinger:
[...]

PR #442, opened 2018-05-22
map2alm pixel weights

zonca:
Implementation of #356.

* Automatically download pixel weights file the first time `map2alm` is used with a specific nside using `astropy.data`
* Apply them before the call to the `map2alm` transform

### DIFF
--- a/healpy/src/_sphtools.pyx
+++ b/healpy/src/_sphtools.pyx
@@ -4,6 +4,7 @@
 import numpy as np
 cimport numpy as np
 from libcpp.string cimport string
+from libcpp.vector cimport vector
 from libc.math cimport sqrt, floor, fabs
 cimport libc
 from healpy import npix2nside, nside2npix
@@ -48,6 +49,11 @@ cdef extern from "alm_powspec_tools.h":
 
 cdef extern from "healpix_data_io.h":
     cdef void read_weight_ring (string &dir, int nside, arr[double] &weight)
+    cdef vector[double] read_fullweights_from_fits(string &weightfile, int nside)
+
+cdef extern from "weight_utils.h":
+    cdef void apply_fullweights (Healpix_Map[double] &map,
+            vector[double] &wgt)
 
 DATAPATH = None
 
@@ -176,7 +182,7 @@ def alm2map_spin_healpy(alms, nside, spin, lmax, mmax=None):
     return maps
 
 def map2alm(m, lmax = None, mmax = None, niter = 3, use_weights = False,
-            datapath = None, gal_cut = 0):
+            datapath = None, gal_cut = 0, pixel_weights_filename=None):
     """Computes the alm of a Healpix map.
 
     Parameters
@@ -193,6 +199,8 @@ def map2alm(m, lmax = None, mmax = None, niter = 3, use_weights = False,
       If True, use the ring weighting. Default: False.
     gal_cut : float [degrees]
       pixels at latitude in [-gal_cut;+gal_cut] are not taken into account
+    pixel_weights_filename : str
+      filename of the pixel by pixel weights
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


mreineck:
I'll check once I get home.

With pure C++, the largest deviations from 1 I see in the recovered a_lm are below 1e-9. So I suspect the problem lies in the interface or on the Python side.

zonca:
also updated to the latest healpix trunk

mreineck:
Sorry, I don't manage to get your branch compiled and set up properly... it keeps failing with
```
Sigma is 0.000000 arcmin (0.000000 rad) 
-> fwhm is 0.000000 arcmin
  File "healpy/test/test_pixelweights.py", line 30, in <module>
    a.test_pixelweights()
  File "healpy/test/test_pixelweights.py", line 23, in test_pixelweights
    alm = hp.map2alm(self.m, lmax=self.lmax, use_pixel_weights=True)
  File "/home/martin/codes/zhealpy/healpy/sphtfunc.py", line 167, in map2alm
    package="healpy")
  File "/home/martin/.local/lib/python2.7/site-packages/astropy/utils/data.py", line 636, in get_pkg_data_filename
    timeout=remote_timeout)
  File "/home/martin/.local/lib/python2.7/site-packages/astropy/utils/data.py", line 1074, in download_file
    if url_mirror in url2hash:
  File "/usr/lib/python2.7/shelve.py", line 110, in __contains__
    return key in self.dict
  File "/usr/lib/python2.7/_abcoll.py", line 388, in __contains__
    self[key]
  File "/usr/lib/python2.7/bsddb/__init__.py", line 270, in __getitem__
    return _DeadlockWrap(lambda: self.db[key])  # self.db[key]
  File "/usr/lib/python2.7/bsddb/dbutils.py", line 68, in DeadlockWrap
    return function(*_args, **_kwargs)
  File "/usr/lib/python2.7/bsddb/__init__.py", line 270, in <lambda>
    return _DeadlockWrap(lambda: self.db[key])  # self.db[key]
TypeError: String or Integer object expected for key, unicode found

```

But perhaps the  problem is simply the `apply_fullweights()` call ... this function actually changes its first argument. If a copy is made at the interface between both languages, you will most likely get your unchanged input back. Could you please check for this?

zonca:
I see that the alm values change if I do the weighting or not, so I don't think that this is a problem.
I haven't tested the code in Python 2.7, I'll try now

mreineck:
You need to set iter=0. Then it seems to work.

healpy/src/_sphtools.pyx
```diff
@@ -301,6 +309,17 @@ def map2alm(m, lmax = None, mmax = None, niter = 3, use_weights = False,
     else:
         w_arr.allocAndFill(2 * nside, 1.)
 
+    cdef vector[double] full_weights
+    if pixel_weights_filename is not None:
+
+        full_weights = read_fullweights_from_fits(pixel_weights_filename.encode("UTF-8"), nside)
+        # pixel weighting requires 0 iterations
+        niter = 0
```
zonca: @mreineck do you mean `niter` here?

mreineck: Most likely yes. For my tests I simply called map2alm() in sphtfunc.py with iter=0.

mreineck:
It's because a(l,-m) = (-1)**m conj(a(l,m)) (or something similar).
For m=0 this allows only real-valued a(l,m).

zonca:
@lpsinger Python 2.7 fails on Travis-CI with the same Unicode error inside `astropy` @mreineck reported above, see https://travis-ci.org/healpy/healpy/jobs/382367728

The problem is that I tried with a conda Python 2.7 on my machine and it worked fine....any suggestion?

mreineck:
This is actually a "feature" that worries me... do you really want a healpy functionality which only works when the machine is connected to the internet?

zonca:
@mreineck there is no way around it, weights files are too big to be bundled inside the package.

lpsinger:
> This is actually a "feature" that worries me... do you really want a healpy functionality which only works when the machine is connected to the internet?

Agreed. Why not distribute this with the package?

mreineck:
I can provide them, if you like. I'll start the calculations tomorrow; no idea how long they will take however ...
The size of a weight file is rougly Npix/2 in (bytes).

pllim:
What if... you add `from __future__ import unicode_literals` in your code and try it with Astropy 2.0.6? I think there is something about unicode vs str stuff going on.

zonca:
thanks @pllim, it works now!

@lpsinger would you like to review this PR?

zonca:
so 4096 should be ~100 MB, I think they are too big, I think using `astropy`'s utils.data to retrieve them is fine.

zonca:
add `use_pixel_weights` also to `hp.anafast`

pllim:
FWIW a fix was put in for the next astropy 2.0.7 release but please let us know if you still encounter this error for `astropy>=2.0.7`. I apologize for any inconvenience caused. xref astropy/astropy#7493

zonca:
removed version pinning of `astropy`, now it should get `2.0.7` for Python 2.7

pllim:
Great! All green now.
```
Searching for astropy==2.0.7
Best match: astropy 2.0.7
Adding astropy 2.0.7 to easy-install.pth file
```

Issue #445, opened 2018-05-28
Standardize default FITS extensions name across HEALPix


zonca:
Agreed with @mreineck and @hivon to set default names of the FITS extensions to:

‘TEMPERATURE’, ‘Q_POLARISATION’, ‘U_POLARISATION’

The same will be used in C++, IDL and F90

zonca:
@mreineck @hivon, in `healpy` we also have:

* `["II", "IQ", "IU", "QQ", "QU", "UU"]` default names when a map array is of length 6, possibly a covariance matrix.
* `COLUMN_n` with `n` from 0 to the number of components for any other number of components

Do you have anything similar somewhere else in HEALPix?

hivon:
Hi Andrea,

> On 28 May 2018, at 15:49, Andrea Zonca <notifications@github.com> wrote:
> 
> @mreineck @hivon, in healpy we also have:
> 
> 	• ["II", "IQ", "IU", "QQ", "QU", "UU"] default names when a map array is of length 6, possibly a covariance matrix.
> 	• COLUMN_n with n from 0 to the number of components for any other number of components
> Do you have anything similar somewhere else in HEALPix?
no, there is no default in that case in F90 or IDL.
It is up to the user to name the columns outside the I, IQU or IQU+derivatives cases.
Note that FITS numbers its column starting at 1 (TTYPE1, …)

Eric

kadrlica:
I'll just point out that these default extension names make very little sense for those of us using `healpy` with optical survey data.

zonca:
thanks @kadrlica how many columns and which names do you generally use for that?

kadrlica:
I don't think that there is a well established convention in optical surveys (at least, not one that spans across scientific collaborations). Within the Dark Energy Survey, healpix maps are generally used to store survey characteristics as a function of sky position (i.e., magnitude limit, coverage fraction, sky brightness, airmass, PSF size, etc.). These are commonly distributed as compressed nside=4096 maps with one map per file (using default EXTNAME `xtension`). These maps usually contain one data column with an additional pixel column when in explicit format.

I know that the gamma-ray community is also using healpix for storing counts maps (and probably also survey characteristics). This is true of both the Fermi Gamma-ray Space Telescope and the Cherenkov Telescope Array. Maybe @cdeil could comment more on the use in `gammapy`.

Maybe someone familiar with the LIGO use case could also weigh in (@lpsinger)?

cdeil:
[...]

PR #446, opened 2018-05-28
Make default FITS extensions names agree with HEALPix

zonca:
Addresses #445


zonca:
change also docstrings

PR #447, opened 2018-05-28
Rotate map

zonca:
A common request is to have functionality to rotate a complete map to another coordinate system.
I needed this for a project and thought I contribute it to `healpy`.

@mreineck @hivon please send me feedback.

I should also add more tests.


zonca:
See IDL code I used to generate the expected output: https://gist.github.com/zonca/401069e1c520e02eaff8cd86149d5900

hivon:
Hi Andrea,

one minor point that I may have misunderstood: 
the FITS file generated in IDL seems to have 3 columns, with the first one set to 0,
while the one read in python and allegedly coming from IDL seems to only have 2 columns.

Cheers,
Eric

zonca:
Thanks Eric, I removed it with Python

Issue #452, opened 2018-06-04
Optional degrees argument for pixelfunc.max_pixrad


kadrlica:
It'd be nice to have an optional argument to return the value from [`pixelfunc.max_pixrad`](https://github.com/healpy/healpy/blob/a391ad9fc3b0a60a1f2603d92d32154d6abfd0c3/healpy/pixelfunc.py#L1345) in degrees rather than radians. This would parallel the call to [`pixelfunc.nside2pixarea`](https://github.com/healpy/healpy/blob/a391ad9fc3b0a60a1f2603d92d32154d6abfd0c3/healpy/pixelfunc.py#L994).

mreineck:
Wouldn't it be cleaner to just say
`np.degrees(pixelfunc.max_pixrad(...))`
This avoids duplication.

kadrlica:
@mreineck how is this "cleaner" and what "duplication" does it avoid? The suggested form is:
```
pixelfunc.max_pixrad(nside, degrees=True)
```

@zonca yes, I can submit a PR.

mreineck:
Well,
`np.degrees(pixelfunc.max_pixrad(nside))`
is actually shorter than
`pixelfunc.max_pixrad(nside, degrees=True)`
and it doesn't require additional code in healpy.

zonca:
the best solution would be to use `astropy.units`, but it would require significant work, I can't commit to this but if somebody volunteers I'd love to review a pull request about that.

mreineck:
I'm sorry if my comments above seem to be unnecessarily destructive ... they definitely were not meant that way.
My idea is that code functionality should be as orthogonal as possible, i.e. if there already is a simple way to achieve something, the code should not be enlarged to support another way of achieving it. In the long run this leads to feature creep and very hard-to-maintain code.
Unfortunately, the `np.degrees()` approach does not apply well to `pixelfunc.nside2pixarea`, I admit.

kadrlica:
@zonca yeah, astropy units would be nice, assuming that the decision has been made to broadly incorporate astropy as a dependency and that it doesn't add much overhead in calculations. Unfortunately, I also don't have time to implement this right now.

@mreineck I don't think your comments were destructive. My main feeling is that this change would preserve parallelism in the API.

kadrlica:
[...]

PR #453, opened 2018-06-04
Optional degrees argument for max_pixrad

kadrlica:
Added optional 'degrees' argument to pixelfunc.max_pixrad (with associated test) to address #452.


zonca:
thanks!

Issue #473, opened 2018-08-02
Build failed with python 3.7


niu541412:
`healpy/src/_query_disc.cpp:9828:21: error: no member named 'exc_type' in '_ts'
    *type = tstate->exc_type;
            ~~~~~~  ^
healpy/src/_query_disc.cpp:9829:22: error: no member named 'exc_value' in '_ts'; did you mean 'curexc_value'?
    *value = tstate->exc_value;
                     ^~~~~~~~~
                     curexc_value
/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/include/python3.7m/pystate.h:237:15: note: 'curexc_value' declared here
    PyObject *curexc_value;
              ^
healpy/src/_query_disc.cpp:9830:19: error: no member named 'exc_traceback' in '_ts'; did you mean 'curexc_traceback'?
    *tb = tstate->exc_traceback;
                  ^~~~~~~~~~~~~
                  curexc_traceback
/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/include/python3.7m/pystate.h:238:15: note: 'curexc_traceback' declared
      here
    PyObject *curexc_traceback;
              ^
healpy/src/_query_disc.cpp:9837:24: error: no member named 'exc_type' in '_ts'
    tmp_type = tstate->exc_type;
               ~~~~~~  ^
healpy/src/_query_disc.cpp:9838:25: error: no member named 'exc_value' in '_ts'; did you mean 'curexc_value'?
    tmp_value = tstate->exc_value;
                        ^~~~~~~~~
                        curexc_value
/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/include/python3.7m/pystate.h:237:15: note: 'curexc_value' declared here
    PyObject *curexc_value;
              ^
healpy/src/_query_disc.cpp:9839:22: error: no member named 'exc_traceback' in '_ts'; did you mean 'curexc_traceback'?
    tmp_tb = tstate->exc_traceback;
                     ^~~~~~~~~~~~~
                     curexc_traceback
/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/include/python3.7m/pystate.h:238:15: note: 'curexc_traceback' declared
      here
    PyObject *curexc_traceback;
              ^
healpy/src/_query_disc.cpp:9840:13: error: no member named 'exc_type' in '_ts'
    tstate->exc_type = type;
[...]

PR #475, opened 2018-08-05
add Python 3.7 to Travis

zonca:
See #473

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ python:
   - 2.7
   - 3.5
   - 3.6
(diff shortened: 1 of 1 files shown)


cdeil:
@zonca - Restart build and see if it works now?

zonca:
ok, there are issues with Anaconda, but it works with the system python 3.7, however we do not want to use the VM based Travis infrastructure, let's wait for 3.7 to land on the default Travis container based platform. However, doing another release with a newer cython should fix this.

Issue #465, opened 2018-07-17
Compilation error on OSX 10.13


jiashuhan:
I tried to upgrade healpy from 1.10 using pip3 install --upgrade healpy, but encountered the following errors:
```
 Installing collected packages: healpy 
 Running setup.py install for healpy ... error
    Complete output from command /Library/Frameworks/Python.framework/Versions/3.5/bin/python3 -u -c "import setuptools, tokenize;__file__='/private/var/folders/bm/pwv0zfm11z1c088dthv9440w0000gn/T/pip-install-lvscuvop/healpy/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /private/var/folders/bm/pwv0zfm11z1c088dthv9440w0000gn/T/pip-record-0g_yalaj/install-record.txt --single-version-externally-managed --compile:
    running install
    running build
    running build_py
    creating build
    creating build/lib.macosx-10.6-intel-3.5
    creating build/lib.macosx-10.6-intel-3.5/healpy
    copying healpy/__init__.py -> build/lib.macosx-10.6-intel-3.5/healpy
    copying healpy/pixelfunc.py -> build/lib.macosx-10.6-intel-3.5/healpy
    copying healpy/sphtfunc.py -> build/lib.macosx-10.6-intel-3.5/healpy
    copying healpy/visufunc.py -> build/lib.macosx-10.6-intel-3.5/healpy
    copying healpy/fitsfunc.py -> build/lib.macosx-10.6-intel-3.5/healpy
    copying healpy/projector.py -> build/lib.macosx-10.6-intel-3.5/healpy
    copying healpy/rotator.py -> build/lib.macosx-10.6-intel-3.5/healpy
    copying healpy/projaxes.py -> build/lib.macosx-10.6-intel-3.5/healpy
    copying healpy/version.py -> build/lib.macosx-10.6-intel-3.5/healpy
    copying healpy/zoomtool.py -> build/lib.macosx-10.6-intel-3.5/healpy
    copying healpy/cookbook.py -> build/lib.macosx-10.6-intel-3.5/healpy
    copying healpy/newvisufunc.py -> build/lib.macosx-10.6-intel-3.5/healpy
    creating build/lib.macosx-10.6-intel-3.5/healpy/test
    copying healpy/test/test_pixelweights.py -> build/lib.macosx-10.6-intel-3.5/healpy/test
[...]

PR #478, opened 2018-08-13
Update minimum healpix-cxx version

lpsinger:
Update the minimum version of healpix-cxx in the pkg-config check. The healpix-cxx API changed in version 3.40.0, and we now require that version to build healpy. If the user has an older version of healpix-cxx installed, then we should use our own bundled sources instead.

This may address #455, #465.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -463,7 +463,7 @@ setup(
         (
             "healpix_cxx",
             {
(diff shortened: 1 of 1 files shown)


zonca:
thanks!

lpsinger:
@zonca, would you please do a new release to pick up this patch?

zonca:
ok, released 1.12.4

Issue #482, opened 2018-08-21
mollview crashes with nest option for plotting


eltrompetero:
Once I set nsides>2, this crashes. I'm using healpy on Python 3.6, the latest version 1.12.3.

`import healpy as hp
from healpy.visufunc import projscatter,mollview

fig,ax=plt.subplots(figsize=(15,4))

m = np.arange(hp.nside2npix(3))
mollview(m,title='',nest=True)`

mreineck:
Hmm, there is no Healpix pixelization with NESTED ordering and Nside=3; for NESTED, Nside needs to be a power of 2. Still, you should get a meaningful exception. Is this not the case?

eltrompetero:
Oh shoot. I was coming back to some old code, and this completely slipped my mind. I might point out that the kernel was dying without any error message being printed to the Jupyter notebook although I see that one is showing up on the terminal.

DanielLenz:
I got the following exception when executing this in ipython, which I'd consider meaningful enough.
Can also confirm that nothing is raised in the notebook itself and the kernels just dies, but the same exception is logged in the terminal.

```python

Error encountered at /Users/travis/miniconda3/conda-bld/healpy_1517058783791/work/healpy-1.11.0/healpixsubmodule/src/cxx/autotools/Healpix_cxx/healpix_base.cc, line 716
(function void T_Healpix_Base<long>::SetNside(I, Healpix_Ordering_Scheme) [I = long])

SetNside: nside must be power of 2 for nested maps

libc++abi.dylib: terminating with uncaught exception of type PlanckError
```

Should we add a call to `pixelfunc.is_nside_ok()` for the different visualization functions in the python layer? This would be redundant, given that it's already caught, but makes for a meaningful and obvious exception in e.g. notebooks.

zonca:
yes, @DanielLenz, I think it is worth it, it makes `healpy` more user friendly

eltrompetero:
Glad to see something useful came out of this!

PR #492, opened 2018-09-17
Add nside tests for all visu funcs, needed for nested maps.

DanielLenz:
Addresses #482 

Nested maps are only defined for nsides that are a power of 2. This is caught, but not in the python layer. This PR allows a graceful exit with a meaningful error in e.g. notebooks and ipython.


mreineck:
Fine with me!

Issue #494, opened 2018-09-19
Adding a Dockerfile?


DanielLenz:
Due to the current issues with OS X builds, I've been doing all the development/testing in a docker container.

Do you think it would add value to have a light-weight [Dockerfile](https://gist.github.com/DanielLenz/868e7544ccbea4fd255be6bb94238f50) as part of the healpy repo? This helps to overcome any installation/dependency issues, allows for easy development, and could e.g. be used to set up an interactive [binder](https://mybinder.org).

Developing/running/testing is then as easy as:

```bash

docker run -d -t -v "$(pwd)":/home/healpy hp:latest
docker exec -it <container_id> /bin/bash
```

zonca:
yes, I think it would be very useful to have, and also have an autobuild on dockerhub

DanielLenz:
While this has been merged, I'll keep it open to see if everything (especially the autobuild) works as intended.

DanielLenz:
Not yet; let me look into the autobuild on dockerhub one more time first.

lpsinger:
Daniel, what are the macOS issues? My main machine is a Mac; I can help.

DanielLenz:
I just tried it out again (it's been a couple of months), and the latest (1.12.10) version installs just fine from source on MacOS with gcc-8. Thanks for offering your help, I'm sure I'll take you up on that offer sooner or later :)

PR #495, opened 2018-09-20
Add Dockerfile

DanielLenz:
Due to the current issues with OS X builds, I've been doing all the development/testing in a docker container. This PR adds a Dockerfile to the healpy repo, some instructions are currently at the beginning of the Dockerfile.

If people are interested, we can also document this elsewhere.

I've also pushed the file to https://hub.docker.com/r/healpy/python3.6-healpy/, please ping me with your dockerhub username to be added to the organization.

Addresses https://github.com/healpy/healpy/issues/494

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,51 @@
+# To use this image, first build the dockerfile by switching
+# to the directory in which it is located and then running
+# >> docker build -t hp .
+# This builds the image, and
+# >> docker images
+# will show you the healpy (hp) image, as well as the
+# underlying python-3.6-slim image.
+#
+# After that, run the image with
+# >> docker run -d -t --name healpy -v "$(pwd)":/home/healpy hp:latest
+# This will run the image (i.e. create a container, which is an instance of 
+# that image) and mount the current directory, so that you can directly modify
+# the content of the container, without having to access it.
+# 
+# You can then use
+# >> docker exec -it healpy /bin/bash
+# to start a bash session in the container, and interact with the healpy package.
+# This is very helpful to install the package, run tests, etc., while the files can be edited
+# and added to source control outside of the container due to the mount.
+
+FROM python:3.7-slim
+
+WORKDIR /home
+
+# Setup tools to work in the container
+RUN apt-get update && apt-get install -y \
+    make \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+    gcc \
+    g++ \
+    git \
+    autoconf \
+    automake \
+    libtool \
+    pkg-config
+
+# Copy the local healpy copy into the docker container
+COPY . /home/healpy
```
zonca: I think it would be useful here to clone master from Github and build that.

zonca:
thanks @DanielLenz , is https://hub.docker.com/r/healpy/python3.6-healpy/ automatically building when we tag a release or make a commit on master?

DanielLenz:
Thanks @zonca , these are great suggestions!

> thanks @DanielLenz , is https://hub.docker.com/r/healpy/python3.6-healpy/ automatically building when we tag a release or make a commit on master?

Not yet! I won't find the time this weekend, but look forward to implementing that as soon as possible.

DanielLenz:
I've just rebased this against `master`, and the [docker repo](https://hub.docker.com/r/healpy/healpy/) is set to auto build, even though I haven't tested that yet.

Could you double check this once the `Dockerfile` is in healpy/master, to see if it works in the live environment?

DanielLenz:
Should we proceed with this, @zonca ? Let me know if you have any more comments or suggestions.

DanielLenz:
I've updated this to python3.7 (had some issues with the installation on the Docker Python 3.6 image), and also changed it to copy the local healpy folder into the docker container. I think the dominant use case for this is development, and the workflow

```bash
> git checkout feature
> docker build ...
> docker run (mounting the local copy)
> <in container> edit, test, install, etc
> exit container, retain all the changes made in it
> git commit/push/merge etc
```

zonca:
I think this is very useful for development
I wonder if we should have instead another version of this for docker hub instead bundles the latest healpy version, so that people which have some environment issues could use a docker container instead of installing healpy directly, we could have a container tagged with the healpy version.
Anyway I will merge this for now.

Issue #436, opened 2018-04-25
Suppress plot when using gnomview


j-erler:
I was wondering if there is a way to suppress the plot that is automatically created when using gnomview. I am only interested in using the projected map array and creating these for a large number of objects slows down my system since a plot is created each time.

zonca:
I think you can move `img = self.proj.projmap(map,vec2pix_func,rot=rot,coord=coord)` outside of `projmap` in `projaxes.py`, then pass `img` as argument to `projmap`. Then add a keyword to `gnomview`, like `create_plot=True` and if `False` just return `img` and exit.

j-erler:
Thank you for the quick reply. I managed to mitigate the problem without modifying the source code by calling plt.clf() and plt.close() after each call of gnomview to delete the image and close the window. This solves the issue of RAM consumption, which was my primary concern.

PR #496, opened 2018-09-21
Add no_plot Boolean variable to gnomview

j-erler:
This commit adds the Boolean no_plot variable to the gnomview function in visufunc.py. Setting the variable to True will close the automatically generated output figure, which is useful is gnomview is run in a loop i.e. when stacking a large number of maps. Without this option, healpy will consume large amounts of RAM and can crash jupyter notebooks.

### DIFF
--- a/healpy/visufunc.py
+++ b/healpy/visufunc.py
@@ -386,9 +386,9 @@ def gnomview(
       Default: None
     notext: bool, optional
       If True: do not add resolution info text. Default=False
-    return_projected_map : bool
+    return_projected_map : bool, optional
       if True returns the projected map in a 2d numpy array
-    no_plot : bool
+    no_plot : bool, optional
       if True no figure will be created      
 
     See Also
@@ -545,7 +545,7 @@ def gnomview(
         if wasinteractive:
             pylab.ion()
             # pylab.show()
-        if no_plot is True:        
+        if no_plot:        
             pylab.close(f)  
             f.clf()
             ax.cla()
--- a/healpy/visufunc.py
+++ b/healpy/visufunc.py
@@ -323,6 +323,7 @@ def gnomview(
     margins=None,
     notext=False,
     return_projected_map=False,
+    no_plot=False,
 ):
     """Plot a healpix map (given as an array) in Gnomonic projection.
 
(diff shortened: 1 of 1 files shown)


healpy/visufunc.py
```diff
@@ -387,6 +388,8 @@ def gnomview(
       If True: do not add resolution info text. Default=False
     return_projected_map : bool
       if True returns the projected map in a 2d numpy array
+    no_plot : bool
```
DanielLenz: Could you make it more consistent and use
`no_plot : bool, optional`?
(and please also add that to the `return_projected_map` keyword)

healpy/visufunc.py
```diff
@@ -542,6 +545,10 @@ def gnomview(
         if wasinteractive:
             pylab.ion()
             # pylab.show()
+        if no_plot is True:
```
DanielLenz: Just use `if no_plot:`

DanielLenz:
Thanks for the PR! Are you aiming to generate many plots and save them without actually showing, or are you just interested in the return map you get from `return_projected_map`?

j-erler:
Thanks for the suggestions on how to make the new code consistent with the previous. I was intending to use `no_plot` together with `return_projected_map` to create a data cube using a for-loop, after which the data could be saved. This requires the least modification to gnomview. Not sure if this feature would be useful to have with the other functions in `visufunc`, since those produce a single image from an all-sky map and I see no reason why you would want to run a loop or suppress the output.  I forgot to mention that this commit solves Issue #436 but is more of a hack than a proper solution since a figure is still created but immediately closed again. Preventing the creation of a figure in the first place would require much deeper modification.

DanielLenz:
Thanks for adjusting that so quickly!

I'm not yet fully getting the other point though, sorry. When talking about a datacube, you mean you have several HEALPix maps at different frequencies/energies/colors, and you want to generate a cube for a small patch of the sky?

If so, why wouldn't that be useful for a different projection, e.g. `cartview`?

DanielLenz:
Thanks for clarifying! We should wait for a few more opinions to see what the extent of this PR should be.

Either way, if you're only interested in reprojecting HEALPix maps to 2D maps, you should check out [`reproject`](https://reproject.readthedocs.io/en/stable/), which is really built for that.

zonca:
thanks @j-erler and @DanielLenz

PR #498, opened 2018-09-26
Change default data type in read_map to np.float32

j-erler:
Saves memory by default. Adds consistency with write_map.

### DIFF
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -264,7 +264,7 @@ def write_map(
 def read_map(
     filename,
     field=0,
-    dtype=np.float64,
+    dtype=np.float32,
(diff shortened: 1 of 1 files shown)


DanielLenz:
Agreed, might not be compatible with existing codes. I also agree that reading/writing is currently inconsistent, so it makes sense to set the default dtype in `hp.write_map()` to `np.float64`.

j-erler:
Thanks for the feedback. I did not consider potential compatibility issues down the line. I would argue however that the default data type in `hp.write_map()` should stay at `np.float32` since there is really no benefit from using double precision is most cases.

zonca:
ok, let's leave it as it is

PR #501, opened 2018-10-15
Update projaxes.py

pqrs6:
Fixes behavior when users pick their own colormap so that by default the background is white and masked pixels are gray.


zonca:
Thanks!

PR #506, opened 2018-11-09
Declare support for Numpy >= 1.13

### LINKED ISSUE #505: Import errors on Python 3.6 and 2.7 but not on 3.4 and 3.5
I help maintain a library that depends on healpy and installs the latest version from PyPI. Recently, with healpy 1.12.4, our builds stopped working but only for Python versions 3.6 and 2.7, and not for 3.4 and 3.5. In these builds, healpy installs without error, but the following import error occurs when healpy is imported for 3.6 and 2.7 (I attach the relevant part of the traceback for 2.7):
https://pastebin.com/1XF7dHqE

I can confirm this doesn't happen for healpy 1.12.3.

lpsinger:
Fixes #505.

After merging this, we should do another release and rebuild the wheels.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -518,7 +518,7 @@ setup(
             "test/data/*.sh",
         ]
     },
-    install_requires=["matplotlib", "numpy", "six", "astropy", "scipy"],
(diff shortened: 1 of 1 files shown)


zonca:
ok, let's target a release next week

PR #508, opened 2018-11-14
Create RELEASE.md

zonca:


lpsinger:
This is good, but it would be even better to get most of this stuff automated using Travis, Appveyor, etc.

PR #510, opened 2018-11-21
Revert "Set random seed to make test_rotator reproducible"

marius311:
Reverts healpy/healpy#487

I'm guessing this wasn't meant to make it into production (at least not in this way), since it means synfast always produces the same map?

CC @lpsinger 

https://xkcd.com/221/ :)


zonca:
thanks @marius311 ...

can you please also set the seed again in the relevant test function?

marius311:
Is it not set already?
https://github.com/healpy/healpy/blob/9805adbe90af7d586319fcaad8a05b28423322bc/healpy/test/test_sphtfunc.py#L127-L128

zonca:
that was `test_sphtfunc`, we needed that in `test_rotator`, testing in #511

PR #518, opened 2018-12-08
fix pix2ang and vec2ang doc string

ecm0:
Hi, 

this pull-request fixes the doc string of pix2ang and vec2ang in pixelfunc module, where
the output angles are given in reverse order.

Cheers,
Eric


zonca:
thanks

PR #523, opened 2018-12-29
Add support for pathlib

DanielLenz:
The preferred way to handle paths in python>=3.4 is by using pathlib.

```python
from pathlib import Path

path = Path('/home/data/')

hpxmap = hp.read_map(
    path.joinpath('myfile.fits')
)
```

If we try to use hp.read_map() with a `Path` instance, it is not parsed correctly and an exception is raised:

`TypeError: First argument should be a input_data, HDUList instance, or HDU instance`

This PR addresses this; astropy.io.fits already supports pathlib, I just had to add a `Path` instance to the allowed types for `pf.read()`.


zonca:
Great thanks!

PR #531, opened 2019-01-28
WIP: ADD numpy random state argument to synalm

sukhdeep1989:
When run in parallel, numpy random number generator can give correlated results across different threads. Added an argument to be able to explicitly pass Randomstate to synalm which should hopefully solve the problem.
see:
https://stackoverflow.com/questions/29854398/seeding-random-number-generators-in-parallel-programs

### DIFF
--- a/healpy/sphtfunc.py
+++ b/healpy/sphtfunc.py
@@ -347,7 +347,7 @@ def alm2map(
         return np.array(output)
 
 
-def synalm(cls, lmax=None, mmax=None, new=False, RNG=None, verbose=True):
+def synalm(cls, lmax=None, mmax=None, new=False, rng=None, verbose=True):
     """Generate a set of alm given cl.
     The cl are given as a float array. Corresponding alm are generated.
     If lmax is None, it is assumed lmax=cl.size-1
@@ -370,8 +370,8 @@ def synalm(cls, lmax=None, mmax=None, new=False, RNG=None, verbose=True):
       If False, use the old ordering, ie by row
       (e.g. TT, TE, TB, EE, EB, BB or TT, TE, EE, BB if 4 cl as input).
       
-    RNG: numpy random number generator state. Useful when generating multiple
-       maps in parallel. Typically, RNG = np.random.RandomState(seed)
+    rng: numpy random number generator state. Useful when generating multiple
+       maps in parallel. Typically, `rng=np.random.RandomState(seed)`
        see:
        https://stackoverflow.com/questions/29854398/seeding-random-number-generators-in-parallel-programs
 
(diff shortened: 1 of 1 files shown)


zonca:
thanks! what happens if someone passes `RandomState` with the same `seed` to all threads/processes?

I am wondering if in the docstring we should recommend to run `rng=np.random.RandomState()` so that `seed` is `None`.

sukhdeep1989:
Hi, 

> thanks! what happens if someone passes `RandomState` with the same `seed` to all threads/processes?

I think in this case maps from different threads/processes might be same (or at least correlated). 
 
> I am wondering if in the docstring we should recommend to run `rng=np.random.RandomState()` so that `seed` is `None`.

I believe this is default behavior currently in healpy or if  rng=None with edits in this pull request. In such a case numpy tries to generate a new seed, but these seeds can be correlated among threads, which can create correlated maps. For how seeds are generated in this case, see https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.RandomState.html

I believe the recommendation should be that when running in parallel, one should generate a seed based on something which will be unique among different processes, eg. process id, iteration number, etc.

zonca:
what is the difference between just doing `np.random.seed(a_different_seed_for_process)` before calling `synalm`?

sukhdeep1989:
Hi,
Sorry I missed the notification. `np.random.seed(a_different_seed_for_process)` before calling `synalm` will work same as rng implementation, as long as each thread has its own copy go RNG otherwise it can have undefined/random behavior. I believe rng implementation is better as it ensures that each process/thread gets its own copy of RNG.

zonca:
ok, I think this would be very useful if we find a way to initialize independently the seed or the RNG for the user with no intervention from the user.
If it is on the user to provide an independent initialization for each thread anyway, then I would prefer not to merge this.

rainwoodman:
For reproducibility the user may want to provide a particular seed.

By hiding the logic internally you will never be able to meet that requirement.

zonca:
the user should be able to provide a seed, if that seed is None, I want to
automatically have independent seeds for each process.

rainwoodman:
The algorithm for deciding independent seed for each process will depend on how processes / threads are organized -- other words it will be very likely platform dependent.

The alternative is to leave the platform dependency to the users. e.g
```
seed = comm.rank
synalm(...., seed=seed)
```
or 
```
seed = os.getpid()
synalm(..., seed=seed)
```
or
```
seed = rng.randint(200000, size=comm.size)[comm.rank]
synalm(..., seed=seed)
```

I do agree a deficit of the current PR is that it doesn't take a seed but takes a RNG. I think it shall take both -- if a integer is passed in, a RNG shall be created with that seed automatically.

PR #535, opened 2019-02-11
How to update the HEALPix C++ library

zonca:
Also includes a policy discussed with the HEALPix team


zonca:
Have not tested the commands but next time I need to actually update HEALPix C++ I'll do that.

It also includes a policy I've discussed with Kris and @hivon.

@mreineck @lpsinger @DanielLenz please let me know if you have any feedback

Issue #538, opened 2019-03-06
Tutorial file location error


harpolea:
In the section 'Reading and writing maps to file' of the tutorial, it says to read a map from the file `../healpy/test/data/wmap_band_imap_r9_7yr_W_v4.fits`. Before this part, there should be an instruction to move to the healpy directory (or somewhere where this relative path will work), or the file location should be changed to an absolute path. 

I also cannot find that exact file - using `wmap_band_iqumap_r9_7yr_W_v4_udgraded32.fits` instead does not give me the same output as shown in the Visualization section of the tutorial (it looks like it's lower resolution?). The same goes for the file in the section on masked maps - I think this should be `wmap_temperature_analysis_mask_r9_7yr_v4_udgraded32.fits`.

DanielLenz:
Thanks for pointing this out! Interestingly, I also couldn't find the file in the history, the following returns nothing:

    git log --all --full-history -- healpy/test/data/wmap_band_imap_r9_7yr_W_v4.fits

It appears though that the smaller, lower resolution files were added in 2014:

    git log --all --full-history -- healpy/test/data/wmap_band_iqumap_r9_7yr_W_v4_udgraded32.fits
```
commit 15d95e6a87a71e3c605c2e8925860dfe7cdb141c
Author: Andrea Zonca <code@andreazonca.com>
Date:   Fri Jan 10 10:30:57 2014 -0800

    Refactoring of the Spherical Armonics tests

    Now nside32 test maps are included in the repository
```

This also explains the issue described in https://github.com/healpy/healpy/issues/539. Estimating the power spectrum up to lmax=1024 with a map at nside=32 won't work, a conservative upper limit for lmax is 2*nside.

My guess would be that the higher resolution files were remove because of their size (~40MB for IQU at nside512). @zonca , what do you think? Add the higher resolution WMAP data again, or update the tutorial with the nside32 map/simulated data?

zonca:
I think at some point we lost the script to download the high resolution WMAP maps, I recovered it from an ancient healpy version, see https://github.com/healpy/healpy/blob/master/healpy/test/data/get_wmap_maps.sh
We should tell people to run that before running the tutorial so they have some real data, or just rework the tutorial at lower resolution, whatever you think is best.

zonca:
@DanielLenz any update on fixing the tutorial?

DanielLenz:
Haven't had time yet, will do it before the weekend.

zonca:
fixed in #543

PR #543, opened 2019-03-15
Add download instructions for high-res WMAP data

DanielLenz:
Minimal solution to address https://github.com/healpy/healpy/issues/538.

I'd furthermore prefer to have a function that loads test data and puts them into the package directory for future use. One example for this is implemented e.g. in [sklearn.datasets.fetch_datasets](https://scikit-learn.org/stable/datasets/index.html#real-world-datasets).

Feel free to comment/merge as needed for JOSS, I'd still like to extend/update the tutorial in the near future.


zonca:
thanks!

Issue #536, opened 2019-02-23
sphtfunc.pixwin returns array of size greater than default lmax=3*nside


msyriac:
This probably just needs clarification or an API extension, but see

```
>>> import healpy as hp
>>> import numpy as np
>>> nside = 128
>>> npix = hp.nside2npix(nside)
>>> hmap = np.random.random((npix,))
>>> alm = hp.map2alm(hmap)
>>> cl = hp.alm2cl(alm)
>>> cl.shape
(384,)
>>> hp.pixwin(nside).shape
(513,)
```

i.e., map2alm defaults to lmax=3 * nside, while pixwin presumably defaults to lmax=4 * nside. Is that correct? If so, it would help to clarify this in the docs, and also possibly extend the pixwin API so that it returns the pixel window up to a desired lmax.

DanielLenz:
Thanks for raising this issue! I support the idea of having an `lmax` keyword, default should be either 3*nside or whatever lmax the pixwins are calculated up to.

If there are no objections, I'll add a PR for this. @zonca

mreineck:
I've probably mentioned this in the past; if so, sorry for the repetition...
Giving a default lmax for `map2alm` is a bad idea. Choosing a good value requires knowledge about the map to be analyzed. In most cases I'd recommend `lmax=2*nside`, values should only ever be higher if the user is sure that the map is properly band-limited (which is virtually never the case, unless you just created the map using `alm2map`). Values above `3*nside` definitely make no sense; I have no idea why the pixel window values go further up.

zonca:
@DanielLenz I think it should be the same as `map2alm`

@mreineck do you have a reference that explains how you are supposed to choose `lmax`? It would be useful to link it or give a summary in the `map2alm` docs.

mreineck:
The most thorough user guide for harmonic analysis on Healpix maps is probably given in the `anafast` documentation from the Fortran package: https://healpix.sourceforge.io/html/fac_anafast.htm

DanielLenz:
@zonca Okay, I'll just use the following:
```
if lmax is None:
        lmax_ = 3 * nside - 1
    else:
        lmax_ = lmax

return pw_temp[:lmax_]
```

zonca:
@DanielLenz please crosscheck but I think it might be `[:lmax+1]`

zonca:
closed by #544

PR #544, opened 2019-03-15
Pixwin lmax

DanielLenz:
Add an `lmax` keyword for `hp.pixwin()`, as well as a couple of simple tests in `test_sphtfunc()`.

Addresses https://github.com/healpy/healpy/issues/536

Ran black formatting afterwards, hence the size of the PR.

### DIFF
--- a/healpy/sphtfunc.py
+++ b/healpy/sphtfunc.py
@@ -965,7 +965,6 @@ def pixwin(nside, pol=False, lmax=None):
       temperature and polarisation pixel window functions.
     """
 
-    # Have lmax default to 3*nside
     if lmax is None:
         lmax = 3 * nside - 1
 
--- a/healpy/sphtfunc.py
+++ b/healpy/sphtfunc.py
@@ -967,7 +967,7 @@ def pixwin(nside, pol=False, lmax=None):
 
     # Have lmax default to 3*nside
     if lmax is None:
-        lmax = 3 * nside
+        lmax = 3 * nside - 1
 
     datapath = DATAPATH
     if not pixelfunc.isnsideok(nside):
@@ -979,9 +979,9 @@ def pixwin(nside, pol=False, lmax=None):
     pw = pf.getdata(fname)
     pw_temp, pw_pol = pw.field(0), pw.field(1)
     if pol:
-        return pw_temp[:lmax], pw_pol[:lmax]
+        return pw_temp[: lmax + 1], pw_pol[: lmax + 1]
     else:
-        return pw_temp[:lmax]
+        return pw_temp[: lmax + 1]
 
 
 def alm2map_der1(alm, nside, lmax=None, mmax=None):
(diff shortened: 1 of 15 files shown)


healpy/sphtfunc.py
```diff
@@ -955,13 +955,20 @@ def pixwin(nside, pol=False):
       The nside for which to return the pixel window function
     pol : bool, optional
       If True, return also the polar pixel window. Default: False
+    lmax : int, optional
+        Maximum l of the power spectrum (default: 3*nside-1)
```
zonca: here says the default is 3*nside-1, below it is 3*nside

healpy/sphtfunc.py
```diff
@@ -972,9 +979,9 @@ def pixwin(nside, pol=False):
     pw = pf.getdata(fname)
     pw_temp, pw_pol = pw.field(0), pw.field(1)
     if pol:
-        return pw_temp, pw_pol
+        return pw_temp[:lmax], pw_pol[:lmax]
     else:
-        return pw_temp
+        return pw_temp[:lmax]
```
zonca: shouldn't this be `[:lmax+1]`?

DanielLenz: You're right - I made these two changes (3*nside instead of 3*nside-1, and [:lmax] instead of [:lmax+1]) to that they'd cancel out.

This doesn't work of `lmax` is actually provided, so let me push a fix.

DanielLenz:
Should be good now @zonca

healpy/sphtfunc.py
```diff
@@ -955,13 +955,20 @@ def pixwin(nside, pol=False):
...
+        Maximum l of the power spectrum (default: 3*nside-1) 
 
     Returns
     -------
     pw or pwT,pwP : array or tuple of 2 arrays
       The temperature pixel window function, or a tuple with both
       temperature and polarisation pixel window functions.
     """
+
+    # Have lmax default to 3*nside
```
zonca: can you remove this comment?

healpy/sphtfunc.py
```diff
@@ -972,9 +979,9 @@ def pixwin(nside, pol=False):
     pw = pf.getdata(fname)
     pw_temp, pw_pol = pw.field(0), pw.field(1)
     if pol:
-        return pw_temp, pw_pol
+        return pw_temp[: lmax + 1], pw_pol[: lmax + 1]
```
zonca: the spacing here looks strange, is `black` ok with this?

healpy/sphtfunc.py
DanielLenz: Done!

healpy/sphtfunc.py
DanielLenz: Yes, this is black's default behavior. This is cleary different to what PEP8 says (https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements).

PR #551, opened 2019-04-22
Allow user to set the background and UNSEEN colors in calls to *view()

keskitalo:
Users can set the color map used to plot the Healpix maps but they have no way to adjust the way the background and UNSEEN pixels are plotted.

This pull request adds two keyword parameters to mollview, orthview, gnomview and azeqview:
* badcolor to set the value used to plot bad values
* bgcolor to set the value of the background
both of these values default to the current library defaults: "gray" for UNSEEN values and "white" for background.

### DIFF
--- a/healpy/projaxes.py
+++ b/healpy/projaxes.py
@@ -139,6 +139,8 @@ class SphericalProjAxes(matplotlib.axes.Axes):
         vmin=None,
         vmax=None,
         badval=UNSEEN,
+        badcolor="gray",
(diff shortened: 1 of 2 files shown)


zonca:
Very nice, I think this is a great usability improvement compared to having the user create a colormap themselves. @daniellenz any feedback?

DanielLenz:
That's a fantastic feature enhancement, thanks for taking the time to work on this.

PR #554, opened 2019-05-22
maintain the FITS dtype in read_map

zonca:
read_map forces input to a specific dtype, by default double.
Now if dtype is None, it maintains the original dtype.

### DIFF
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -294,6 +294,7 @@ def read_map(
       Force the conversion to some type. Passing a list allows different
       types for each field. In that case, the length of the list must
       correspond to the length of the field parameter. Default: np.float64
+      if None, keeps the same dtype of the input FITS file
     nest : bool, optional
       If True return the map in NEST ordering, otherwise in RING ordering;
       use fits keyword ORDERING to decide whether conversion is needed or not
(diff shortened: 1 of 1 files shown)


healpy/fitsfunc.py
```diff
@@ -397,15 +398,21 @@ def read_map(
 
     for ff, curr_dtype in zip(field, dtype):
         try:
```
DanielLenz: Is the try/except statement statement still necessary? Perhaps we can simply call `hdu.verify("fix")`, and then proceed to read?

On a related note, it would also be good to gradually move away from print statements and use the `logging` module or (even better suited for users) `warnings.warn()`.

healpy/fitsfunc.py
```diff
@@ -294,6 +294,7 @@ def read_map(
       Force the conversion to some type. Passing a list allows different
       types for each field. In that case, the length of the list must
       correspond to the length of the field parameter. Default: np.float64
+      if None, keeps the same dtype of the input FITS file
```
DanielLenz: If None, keep the dtype of the input FITS file

healpy/fitsfunc.py
zonca: I prefer to use the try/except, not sure how expensive is the verification step.

yes, definitely that would be great to remove all the prints, probably `warnings` is enough for our purposes

DanielLenz:
Thanks, feel free to merge after travis has run!

PR #557, opened 2019-05-27
Update HEALPix to 3.60

zonca:
The HEALPix team is about to release version 3.60, updating it within `healpy`.


zonca:
@mreineck when I update to healpix 3.60, I get this error in the configuration:

```
checking for ffgnrwll in -lcfitsio... no
configure: error: could not find the cfitsio library
```

This happens both with cfitsio 3.45 and with the latest 3.47, and I am running this outside of `healpy` for testing.

I installed `cfitsio` and then ran `autoreconf -i` and then `./configure` from the `src/cxx` folder.

mreineck:
Where is the libcfitsio installed? Which platform?

The output you show seems to indicate that `configure` has found the the `cfitsio` header; that should be on the line before `checking for ffgnrwll in -lcfitsio`. In my libcfitsio (3.45 coming with debian testing) the symbol is present.

zonca:
yes, correct, it found it,

and the header has that symbol:

```
/home/zonca/anaconda/envs/pysm2/include/fitsio.h:int CFITS_API ffgnrwll(fitsfile *fptr, LONGLONG *nrows, int *status);   
```

I am under ubuntu and installing `cfitsio` myself into a conda environment.

mreineck:
If the library is installed in a nonstandard location, you need to tell `configure` where it is.

If you have `pkgconfig` info for  libcfitsio, it should be sufficient to set the PKG_CONFIG and PKG_CONFIG_PATH environment variables accordingly, otherwise you probably have to set CFITSIO_CFLAGS and CFITSIO_LIBS. `./configure --help` might provide some useful information here.

zonca:
it looks like `pkg-config` works fine:

```
pkg-config --cflags --libs cfitsio
-I/home/zonca/anaconda/envs/pysm2/include -L/home/zonca/anaconda/envs/pysm2/lib -lcfitsio
```

mreineck:
Hmm, then I don't know what's going on ... @lpsinger, any idea?

zonca:
maybe something in the code that checks for the symbol?

```
save_CFLAGS=$CFLAGS                                                                                                                                                        
save_LIBS=$LIBS                                                                                                                                                            
CFLAGS="$CFLAGS $CFITSIO_CFLAGS"                                                                                                                                           
LIBS="$LIBS $CFITSIO_LIBS"                                                                                                                                                 
{ $as_echo "$as_me:${as_lineno-$LINENO}: checking for ffgnrwll in -lcfitsio" >&5                                                                                           
$as_echo_n "checking for ffgnrwll in -lcfitsio... " >&6; }                                                                                                                 
if ${ac_cv_lib_cfitsio_ffgnrwll+:} false; then :                                                                                                                           
  $as_echo_n "(cached) " >&6                                                                                                                                               
else                                                                                                                                                                       
  ac_check_lib_save_LIBS=$LIBS                                                                                                                                             
LIBS="-lcfitsio  $LIBS"                                                                                                                                                    
cat confdefs.h - <<_ACEOF >conftest.$ac_ext                                                                                                                                
/* end confdefs.h.  */                                                                                                                                                     
                                                                                                                                                                           
/* Override any GCC internal prototype to avoid an error.                                                                                                                  
   Use char because int might match the return type of a GCC                                                                                                               
   builtin and then its argument prototype would still apply.  */                                                                                                          
#ifdef __cplusplus                                                                                                                                                         
extern "C"                                                                                                                                                                 
#endif                                                                                                                                                                     
char ffgnrwll ();                                                                                                                                                          
int                                                                                                                                                                        
main ()                                                                                                                                                                    
{                                                                                                                                                                          
return ffgnrwll ();                                                                                                                                                        
  ;                                                                                                                                                                        
  return 0;                                                                                                                                                                
}                                                                                                                                                                          
_ACEOF                                                                                                                                                                     
if ac_fn_c_try_link "$LINENO"; then :                                                                                                                                      
  ac_cv_lib_cfitsio_ffgnrwll=yes                                                                                                                                           
else                                                                                                                                                                       
  ac_cv_lib_cfitsio_ffgnrwll=no                                                                                                                                            
fi                                                                                                                                                                         
rm -f core conftest.err conftest.$ac_objext \                                                                                                                              
    conftest$ac_exeext conftest.$ac_ext                                                                                                                                    
LIBS=$ac_check_lib_save_LIBS                                                                                                                                               
fi                                                                                                                                                                         
{ $as_echo "$as_me:${as_lineno-$LINENO}: result: $ac_cv_lib_cfitsio_ffgnrwll" >&5                                                                                          
$as_echo "$ac_cv_lib_cfitsio_ffgnrwll" >&6; }                                                                                                                              
if test "x$ac_cv_lib_cfitsio_ffgnrwll" = xyes; then :                                                                                                                      
  :                                                                                                                                                                        
else                                                                                                                                                                       
  as_fn_error $? "could not find the cfitsio library" "$LINENO" 5                                                                                                          
fi   
```

mreineck:
This code is autogenerated, so there is not much I can do about it.
Also, we had pratically the same code in the standalone Healpix C++ package for years...

zonca:
I think it is a problem with the compiler embedded in conda, if I use the Ubuntu compilers and Ubuntu cfitsio, it passes the `cfitsio` tests.
I would have preferred to use the conda compilers (https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html) because are how most packages in conda are built, but I didn't find a way.

Next, I get:

```
checking for pkg-config... /usr/bin/pkg-config
checking pkg-config is at least version 0.9.0... yes
checking for cfitsio... yes
checking for ffgnrwll in -lcfitsio... yes
checking fitsio.h usability... yes
checking fitsio.h presence... yes
checking for fitsio.h... yes
checking for libsharp... no
checking for sharp_execute in -lsharp... no
configure: error: could not find the libsharp library
```

do I need to install `libsharp` separately?

mreineck:
> do I need to install libsharp separately?

Yes, the current philosophy is that you need to install `libsharp` and then `Healpix_cxx`. `libsharp` was separated, because the Fortran Healpix library needs it as well.
But the sources are all in the repo, under `src/common_libraries`.

Be sure to have a look at libsharp's `COMPILATION` file to determine the best set of flags for your purpose. I expect that `-DMULTIARCH -std=c99 -O3 -ffast-math -fopenmp` should be best suited for healpy distribution.

zonca:
implemented in #589

PR #581, opened 2019-10-25
Add dist2holes function

xgarrido:
This PR should fix issue #529 at the level of ```healpy``` but needs additionnal files in ```Healpix/C++``` library

- ```mask_tools.h``` : header file holding the declaration of functions used in C++ code
- ```mask_tools.cc``` : source code of ```dist2holes``` function (almost a copy-paste of Fortran code from E. Hivon)
- ```Makefile.am``` : the ```autotools``` Makefile needed by ```healpy/healpixsubmodule``` to compile the above files within healpy

The three files are provided as a ```gist``` https://gist.github.com/xgarrido/496847450b0e25a5325b4d79bd9907c8. Feel free to modify/add them to ```Healpix/C++``` library.


xgarrido:
Of course ```Travis``` integration fails until the C++ files will be added to ```Healpix/C++``` library...

mreineck:
Thanks!
I will have a closer look at the C++ code as soon as I can (sorry, pretty busy at the moment), but I can already see one obvious problem:
```
    mask_out.neighbors(p, nb);
    for (size_t i = 0; i < 8; i++) {
      if (mask_out[nb[i]] == PIXGOOD) {
  [...]
```
You cannot unconditionally subscribe `mask_out` with the contents of `nb`, since some neighbours may not exist, in which case the corresponding value in `nb` will be -1, potentially causing a segfault.

xgarrido:
Right I will fix it.

mreineck:
I have an idea for a distance algorithm that should be both simpler and quicker than the current one, but I'll need a few days for testing and tweaking.

amaurea:
@mreineck Is your idea to keep visiting the neighbors of the points with the lowest distance so far? I used two variants of this (one with a heap to visit them strictly in order of distance, and one avoiding a heap which visits pixels in a growing tree-ring-like way) for the distance transform in `pixell`. Operating on the same number of pixels as a healpix nside 4096 map (but using enmaps which have a pixelization where it's easier to find neighbors) this takes about 46 seconds on one core. But the algorithm isn't very parallelizable. I started implementing it for healpix, but since I don't want to link in healpix as a dependency, writing up the neighborhood lookup logic for healpix manually is very tedious.

One good thing about both these methods is that it's simple to add support for a max_dist parameter that makes the search stop at a certain distance. Very often one only wants distances up to a certain point for apodization purposes, and then it's a waste to spend most of the time on distances greater than that. For the same nside 4096-equivalent case I got a run-time of 4 s with max_dist = 1 degree, for example. I recommend adding this parameter to healpy if the algorithm you end up using can benefit from it.

Another nice thing about implementing this stuff is that a general distance_to function comes basically for free, letting you get the distance from each pixel to an arbitrary set of points that are not necessarily at pixel centers.

mreineck:
I'm trying to implement this in a recursive way, see the experimental implementation below:

```
#include <vector>
#include <functional>
#include "Healpix_cxx/healpix_map.h"
#include "geom_utils.h"

using namespace std;
Healpix_Map<double> distcalc (const Healpix_Map<double> &mask, double maxdist)
  {
  int maxord = mask.Order();
  planck_assert(maxord>=0, "Nside must be a power of 2");
  vector<Healpix_Map<bool>> omask; // masks at orders [0; maxord]
  vector<double> buf; // 2*max_pixrad at prders [0; maxord]
  for (int i=0; i<=maxord; ++i)
    {
    omask.emplace_back(i, NEST);
    buf.push_back(2*omask[i].max_pixrad());
    }

  if (mask.Scheme()==RING)
    for (int i=0; i<mask.Npix(); ++i)
      omask[maxord][i] = mask[mask.nest2ring(i)]!=0;
  else
    for (int i=0; i<mask.Npix(); ++i)
      omask[maxord][i] = mask[i]!=0;

  for (int j=maxord-1; j>=0; --j)
    for (int i=0; i<omask[j].Npix(); ++i)
      {
      omask[j][i] = omask[j+1][i*4  ] || omask[j+1][i*4+1]
                 || omask[j+1][i*4+2] || omask[j+1][i*4+3];
                 }

  Healpix_Map<double> tdist(maxord, NEST);
  tdist.fill(maxdist);
  function<void(int, int, const vector<int> &)> process;
  process = [&omask,&tdist,maxord,&process,&buf,maxdist](int order, int pix, const vector<int> &submask)->void
    {
    if (submask.empty()) return;
    vec3 v=omask[order].pix2vec(pix);
    if (order==maxord)
      {
      double d=maxdist;
      for (auto idx:submask)
        d=min(d, v_angle(tdist.pix2vec(idx), v));
      tdist[pix]=d;
      return;
      }
    vector<double> subdist;
    subdist.reserve(submask.size());
    double dlim=1e8;
    for (size_t i=0; i<submask.size(); ++i)
      {
      subdist[i] = v_angle(v, omask[order].pix2vec(submask[i]));
      dlim = min(dlim,subdist[i]);
      }
    if (dlim>maxdist+buf[order]) return;

    // not yet full order;
    vector<int> submask2;
    submask2.reserve(4*submask.size());
    for (size_t i=0; i<submask.size(); ++i)
      if (subdist[i]<dlim+2*buf[order])
        for (int j=submask[i]<<2; j<(submask[i]+1)<<2; ++j)
          if (omask[order+1][j]) submask2.push_back(j);
    for (int pix2=pix<<2; pix2<(pix+1)<<2; ++pix2)
      process(order+1, pix2, submask2);
    };

  vector<int> submask;
  for (int i=0; i<12; ++i)
    if (omask[0][i]) submask.push_back(i);
#pragma omp parallel for schedule(dynamic,1)
  for (size_t i=0; i<12; ++i)
    process(0, i, submask);
  if (mask.Scheme()==RING) tdist.swap_scheme();
  return tdist;
  }
#include "lsconstants.h"
#include "Healpix_cxx/healpix_map_fitsio.h"

int main()
  {
  Healpix_Map<double> bla(12,RING);
  bla.fill(0.);
  for (int i=0; i<bla.Npix(); i+=123456) bla[i]=1.;
//  bla[bla.ang2pix(pointing(pi,0))]=1.;
  auto res=distcalc(bla,pi);
  write_Healpix_map_to_fits("!bla.fits",res,PLANCK_FLOAT32);
  }
```

mreineck:
It seems that there are cases where either of the algorithms runs far too slowly. I'm currently testing with a mask that covers the northern hemisphere (not terribly far-fetched) and with Nside=2048. @xgarrido's port of the Fortran algorithm takes around 100 seconds on my laptop to compute the distances, and run time appears to scale worse than quadratically with Nside. My approach is even worse.

Sorry, but there must be a better way to do this.
@amaurea, how do you compute the exact distances to the mask in your approach when you reach a new pixel via successive `neighbour` calls? You somehow still need to know which pixel of the mask is the closest to this new pixel.

amaurea:
@mreineck 

1. Make a pass over the mask, finding its border, resulting in a set of border pixels and their coordinates.
2. create two maps: a distance map which is initialized to a large number (inf or 1e300 or something) but which eventually contain the distance of each pixel to the closest point; and a domain map that is initialized to -1 but will eventually contain the index of the point that pixel was closest to.
3. Loop through the set of border pixels from #1. Set the distance map to their distance from the corresponding border point, which will be zero in this case, and the domain map to the index of each such point. Add each pixel updated this way to a distance-minimum-heap.
4. Get pixel corresponding to the top of the heap and remove it from the heap, and read off the domain value at its location. That's the id of the point that that pixel was closest to. Loop through the pixel's neighbors.  If their domain is equal to our domain, skip them. Otherwise calculate the distance from our point to them. If that distance is smaller than the existing distance, update the distance and domain for the pixel with the new distance and our domain. Then add each pixel changed this way to the heap.
5. Repeat #4 until the heap is empty.

The result is a distance map and also a nice domain map that says which point in the original set of points each pixel was closest to. For a plain distance transform this isn't that useful, but the function can also be used as a plain "get the distance of each pixel from an arbitrary set of points", and in that case it can be useful, effectively giving a pixelized voronoi diagram. I also use it for a variant of the distance transform called the labeled distance transform, where the input is not just a boolean mask, but a set of integer-marked regions. In that case one can use the border-point-domains that the algorithm outputs to produce the label-domains. When doing the distance transform one should also remember set the distance in the interior of the mask to 0, which can be done simply by multiplying by the mask in the end.

The algorithm is not guaranteed to be accurate to floating point precision. It relies on there being an unbroaken chain of pixels-that-are-closest-to-point-x-on-the-border, but in some circumstances this can fail, in which case the distance will be to the neighbor of that point instead. This should be rare, and the error made will be small.

The algorithm does not rely on the use of a heap for correctness. One can process the points in the working set in any order - the heap is just the choice that leads to the fewest visits to each pixel in the map. An alternative to a heap is to have two working sets, making a linear pass through the first and adding each updated pixel to the second, and then switching their roles for the next pass. That was my original implementation, and it works better for scattered points. I call this the "bubble" version of the algorithm.

Since implementing all of this I have also thought of ways to make the algorithm memory-local and paralellizable, but it would be a bit involved and I haven't implemented it, so I'll just sketch it here:

1. Split the map into tiles, which each tile being a copy of the region in the map for memory locality purposes.
2. Run the algorithm above on each tile for N steps, starting from the points in each tile, and pausing any pixels in the working set that reaches the tile border. This can be done independently for each tile.
3. After those N steps, exchange working set pixels that are at the border with the tiles across that border.
4. Then run another N steps, etc., until all the working sets are empty.
5. Copy the distances and domains for the tiles back into full-sized maps.

If using the bubble method, which I think will be best here, one can even adjust N per tile to let tiles that have fallen behind in distance covered (tiles in area which higher pixel density, for example, though that won't happen in healpix, or tiles in a direction that's harder to travel by neighbor-hopping) catch up. That way the bubble method can approximate the optimality of the heap method without actually having to use a heap.

I wrote a heap-based version for healpix as a bonus in my pixell distance transform. You can see it in `distance_from_points_heap_healpix` in https://github.com/simonsobs/pixell/blob/master/cython/distances_core.c

mreineck:
Thanks a lot fir this beautiful explanation, @amaurea! There are a lot of details and neat tricks I could have never figured out on my own.
For starters, I'll have a go at the standard serial implementation. It's true that the Healpix neighbor function is not terribly fast, but I expect that the distance calculations may actually dominate the runtime.

mreineck:
The algorithm is really beautiful, and it will be very efficient as long as `maxdist` is small: 

```
Healpix_Map<double> distcalc_v3 (const Healpix_Map<double> &mask, double maxdist)
  {
  Healpix_Map<double> res(mask.Nside(), mask.Scheme(), SET_NSIDE);
  Healpix_Map<int> dom(mask.Nside(), mask.Scheme(), SET_NSIDE);
  struct HeapEntry
    {
    vec3 pos;
    double dist;
    int pix;
    HeapEntry(const vec3 pos_, double dist_, int pix_)
      : pos(pos_), dist(dist_), pix(pix_) {}
    bool operator<(const HeapEntry &other) const
      { return dist>other.dist; }
    };
  priority_queue<HeapEntry> heap;
  for (int i=0; i<mask.Npix(); ++i)
    {
    if (mask[i]==0.)  // inside the mask
      {
      dom[i]=i;
      res[i]=0.;
      fix_arr<int,8> nb;
      mask.neighbors(i,nb);
      bool border=false;
      for (size_t i=0; i<8; ++i)
        if (nb[i]>=0) border |= (mask[nb[i]]!=0);
      if (border)
        heap.push(HeapEntry(mask.pix2vec(i), 0., i));
      }
    else  // outside
      {
      dom[i]=-1;
      res[i]=maxdist;
      }
    }

  while(!heap.empty())
    {
    auto entry=heap.top();
    heap.pop();
    fix_arr<int,8> nb;
    mask.neighbors(entry.pix,nb);
    for (size_t i=0; i<8; ++i)
      if ((nb[i]>=0) && (res[nb[i]]>0) && (dom[nb[i]]!=dom[entry.pix]))
        {
        auto pos = mask.pix2vec(nb[i]);
        auto dist = v_angle(pos, entry.pos);
        if (dist<res[nb[i]])
          {
          res[nb[i]]=dist;
          dom[nb[i]]=dom[entry.pix];
          heap.push(HeapEntry(entry.pos, dist, nb[i]));
          }
        }
    }
  return res;
  }
```

mreineck:
Would it be acceptable to input a `Healpix_Map<bool>` instead of a `Healpix_Map<double>`? As it is, the original C++ implementation above only allowed the discrete values 0 and 1 in the input anyway.

xgarrido:
I did some tests given the following "physical" mask (NSIDE = 4096)

mask

on my old laptop (4 CPU). Here are the results of ```time``` command:

- ```dist2holes/F90``` : 723,56s user 4,92s system 353% cpu 3:26,32 total
- ```dist2holes/C++``` : 717,59s user 3,78s system 317% cpu 3:47,31 total
- ```distcalc_v5``` with ```maxdist=pi``` : 662,08s user 2,72s system 314% cpu 3:31,39 total

I set ```maxdist``` to maximum pi value to make the ```distcalc_v5``` function works like ```dist2holes``` program. Your code is faster and, of course, the final result i.e. the distances are all the same.

I also try your code with a smaller ```maxdist=1.0``` given that the maximum distance is close to 0.68 radians. Surprisingly, the gain is almost zero (661,81s user 2,86s system 346% cpu 3:11,80 total).

As I wrote, the ```dist2holes/C++``` code is almost a copy/paste of the original fortran code by @hivon and I think even a fortran coder can read it and understand it as a pure translation of the original code. Your code is faster but as @thibautlouis mentions, the time is not so crucial (of course, having a faster code is always better). On the other hand, you are the main developer of C++ port of Healpix and if you are more confortable with your implementation and think it will be easier to maintain, it is ok for me. What we would like is to have this code asap in ```healpy``` whatever its C+++ implementation is.

xgarrido:
> Would it be acceptable to input a `Healpix_Map<bool>` instead of a `Healpix_Map<double>`? As it is, the original C++ implementation above only allowed the discrete values 0 and 1 in the input anyway.

My first try was to use boolean ```Healpix_Map``` but as far as I understand ```healpy```, it only support ```Healpix_Map<double>``` and I was not able to write the cython part with a boolean map.

mreineck:
Thanks for testing, @xgarrido!
> I also try your code with a smaller maxdist=1.0 given that the maximum distance is close to 0.68 radians. Surprisingly, the gain is almost zero (661,81s user 2,86s system 346% cpu 3:11,80 total).

If the maximum actual distance is 0.68 radians, then setting `maxdist=1.0` should not give any speedup. I'd only expect speedups if `maxdist` is smaller than the maximum real distance.

After working on this code, I now have quite a bit of confidence that it does the right thing in a fairly efficient way and would like to keep it the way it is. I've added it to the Healpix repository, so it should become part of the next release, but I don't know when exactly this will happen.
In the meantime I don't think it should be any problem to add the sources to `healpy`temporarily  (for the case that the next healpy release happens before the next Healpix release.

zonca:
@xgarrido the dist2holes files are now in HEALPix C++, can you rebase your work on master so we can get this into the next release?

Issue #583, opened 2019-11-04
Should query_disc and friends require nside to be a power of two if the scheme is RING?


tmolteno:
Hi There,

Should query_disc and friends require nside to be a power of two if the scheme is RING? I think this restriction only applies to the NESTED scheme for healpix.

I have modified the isnsideok() function to accept non-power of two values, and things seem to work just fine.

PR #584, opened 2019-11-04
Allow non-power-of-two nside values when the pixel scheme is RING

tmolteno:
Hi There,

This PR changes the isnsideok() function to allow non-power-of-two nside values (when using the RING scheme). This is useful when searching for a pixelization scheme that is just high enough - which can lead to huge efficiencies in the situation where one is discretizing a small subset of the sphere at high resolution.

The test is somewhat application specific (to suit my needs) so apologies in advance for that.

Refers to Issue #583.

Tim

### DIFF
--- a/healpy/src/_query_disc.pyx
+++ b/healpy/src/_query_disc.pyx
@@ -50,7 +50,7 @@ def query_disc(nside, vec, radius, inclusive = False, fact = 4, nest = False, np
     are returned, at the cost of increased run time.
     """
     # Check Nside value
-    if not isnsideok(nside):
+    if not isnsideok(nside, nest):
         raise ValueError('Wrong nside value, must be a power of 2, less than 2**30')
     cdef vec3 v = vec3(vec[0], vec[1], vec[2])
     cdef Healpix_Ordering_Scheme scheme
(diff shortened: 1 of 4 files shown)


healpy/src/_query_disc.pyx
```diff
@@ -344,8 +344,11 @@ cdef pixset_to_array(rangeset[int64] &pixset, buff=None):
             ii += 1
     return ipix
 
-cdef bool isnsideok(int nside):
-     return nside > 0 and ((nside & (nside -1))==0)
+cdef bool isnsideok(int nside, bool nest):
```
zonca: what about having `nest` default to False?
That is common in all other functions that accept that argument.

zonca:
also, there is another `isnsideok` method in `pixelfunc.py`...https://github.com/healpy/healpy/blob/master/healpy/pixelfunc.py#L1160
can you remove that so we always use the one defined in cython?

tmolteno:
Hi There. I will make the default argument as you suggest. Good point about the duplicate in pixelfunc. Will submit another PR soon.

tmolteno:
I have left the version in pixelfunc (suitably modified) as it seems a little more logical for isnsideok to be in the pixel functions (happy to shift it if necessary)

zonca:
there is actually another one in `pixelfunc.py` https://github.com/healpy/healpy/blob/master/healpy/pixelfunc.py#L1160, can you remove that one as well?

tmolteno:
Wow, there were three of them! The isnsideok() in https://github.com/healpy/healpy/blob/master/healpy/pixelfunc.py#L1160 is exported to the healpy module. Didn't see it there at all as I had my eyes down looking inside the .pyx files. I will leave that one and remove the one I modified in _pixelfunc.pyx (which has identical functionality, but has much uglier documentation.)

zonca:
Thanks!

Issue #585, opened 2019-11-12
Extremely confusing dtype defaults in fitsfunc


mreineck:
Currently
-  the `dtype` argument of `write_map` defaults to `np.float32`,
-  the one of `read_map` defaults to `np.float64`,
-  the `one of `write_alm` is called `dtype_out` and defaults to `None`,
-  and the one of `read_alm` doesn't exist.

Furthermore, the `map` routines accept lists of types, and the `alm` routines appear to require scalars.

I can prepare a patch, but I don't know what a useful consensus would be. My suggestion is to introduce `None` for all functions and make this the default, but it will change the behaviour of many existing codes.

crosset:
Some bit of history concerning the first two points: the idea was to save space when storing the maps on disk (assuming that usually don’t need more than 7 digits precision on your data) but putting data in a float64 array directly since you usually want to do computation (especially spherical harmonics transform) with double precision. I think I was inspired initially in these defaults by IDL healpix routines.
Of course, it can evolve to something more rational. Critical code should always specify explicitly the data type. A period with a warning (to announce the future change) before making the change could help the transition.

mreineck:
I should also explain why I have become worried about this:

The reason was that a user:
- had a map in double precision
- stored it to disk (using default data types)
- loaded it from disk (again with defaults)
- made sure it was still double precision
- went on to make an accuracy comparison with another code.

This led to the incorrect result that Healpix map analysis performed much worse than the other algorithm, and that in turn was published on arxiv, and we were lucky to get to know about it and point out the mistake. The chance that a referee who is not very familiar with the Healpix software would have spotted this, are remote.

I don't think the user is to blame at any point there.

zonca:
fixed in #586

PR #586, opened 2019-11-12
try to make dtype defaults more consistent

mreineck:
related to #585 

I don't know the procedure how to make compatibility warnings (as @crosset suggested), but can add these as soon as someone has told me :).

### DIFF
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -346,6 +346,14 @@ def read_map(
     m | (m0, m1, ...) [, header] : array or a tuple of arrays, optionally with header appended
       The map(s) read from the file, and the header if *h* is True.
     """
+    # Temporary warning for default dtype
+    if dtype == np.float64:
+        warnings.warn(
+            "If you are not specifying the input dtype and using the default "
+            "np.float64 dtype of read_map(), please consider that it will "
+            "change in a future version to None as to keep the same dtype of "
+            "the input file: please explicitly set the dtype if it is "
+            "important to you.")
 
     fits_hdu = _get_hdu(filename, hdu=hdu, memmap=memmap)
 
@@ -412,9 +420,6 @@ def read_map(
             fits_hdu.verify("fix")
             pix = fits_hdu.data.field(0).astype(int, copy=False).ravel()
 
-    #FIXME: at some point in the future, add:
-    # if dtype is None:
-    #     dtype = [fits_hdu.data.field(ff).dtype for ff in field]
     try:
         assert len(dtype) == len(
             field
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -288,7 +288,7 @@ def write_map(
 def read_map(
     filename,
     field=0,
-    dtype=None,
+    dtype=42,
     nest=False,
     partial=False,
     hdu=1,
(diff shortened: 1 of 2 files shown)


crosset:
# At the top with other imports
import warnings 

# in each function where there is a change (adapt the message)
warnings.warn("The default dtype of write_map() will change in a future version: explicitly set the dtype if it is important to you",
                         category=FutureWarning)

mreineck:
Thanks!

mreineck:
Does this look OK?

crosset:
Yes, it looks good! Thanks!

healpy/fitsfunc.py
```diff
@@ -294,7 +314,9 @@ def read_map(
       Force the conversion to some type. Passing a list allows different
       types for each field. In that case, the length of the list must
       correspond to the length of the field parameter. Default: np.float64
-      if None, keep the dtype of the input FITS file
```
zonca: the issue here is that we are loosing a functionality, which is maintaining the dtype of the input file, which I think is very useful.
If the default is to use `np.float64`, I think it is better to show it into the `read_map` signature directly as it is currently.

I would leave `read_map` as it is and only modify `read_cl` and `write_cl`

mreineck: I think I understand ... but then it becomes really tricky to decide when the FutureWarning should be printed.
In the future all defaults are supposed to change to `None`, of course.

mreineck: OK, I think I found a solution. But it's incredibly ugly...

zonca: yeah, too ugly, I would leave it as it was.
As long as things are documented and do not change defaults from one version to another, I don't think inconsistent is a big issue.

mreineck: Sorry, I have to disagree with this statement. An inconsistency in the interface is always a burden on the user: they have to keep in mind that they need to use some function with special caution, because there is some inconsistency. But on the other hand, default arguments are supposed to make things _easier_ for the user. This is clearly not happening here.

I'm perfectly fine with simply removing all the `dtype` defaults. That would in fact be my favorite option.

Concerning the ugliness: it would of course go away once the switch to the new interface is complete.

zonca: I think we can keep the current behavior *and* print the `FutureWarning`, later on, we will make the default to have `dtype=None` and keep whatever `dtype` is in the file.

mreineck: OK. Does this look better now?

zonca: I am worried people won't see the warning in the docstring.
If think it is useful to check if the input `dtype` was `np.float64` and in that case print a `FutureWarning` that explains that 

> If you are not specifying the input dtype and using the default `np.float64` dtype of read_map(), please consider that it will change in a future version to None as to keep the same dtype of the input file: please explicitly set the dtype if it is important to you

Unfortunately also people that specify `np.float64` will get this warning, but no easy way around it, and better be more verbose then not.

healpy/fitsfunc.py
```diff
@@ -389,6 +412,9 @@ def read_map(
             fits_hdu.verify("fix")
             pix = fits_hdu.data.field(0).astype(int, copy=False).ravel()
 
+    #FIXME: at some point in the future, add:
+    # if dtype is None:
+    #     dtype = [fits_hdu.data.field(ff).dtype for ff in field]
```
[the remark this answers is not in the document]
mreineck: You are right, that has become unnecessary again after the partial reversion I did. Thanks!

healpy/fitsfunc.py
mreineck: Well, that was exactly the reason for the "ugly" default of 42, which would have allowed us to discern between the user explicitly specifying `np.float64` or nothing at all.
Would things be better if I replaced `42` by some opaque dummy value like `healpy.NOT_SET`?

zonca: no I'd rather print an extra warning that people can easily disable than implement some clever workaround

zonca:
thanks!!

PR #589, opened 2019-12-01
Update healpix-cxx to 3.60

### LINKED ISSUE #576: Interface to HEALPix 3.60
@hivon and @mreineck are ready to release HEALPix 3.60.
In the past it has been pretty easy to update the HEALPix C++ code included and built with `healpy` because the interface and the compilation strategy never changed.
Now they have split the HEALPix package compilation stage in 3 packages:

* `pocketfft`
* `libsharp`
* `HEALPix` C++ itself

all 3 of them are compiled with `autotools`.
So we would have to create 2 other Python extensions that take care of building `pocketfft` and `libsharp`, if we are lucky enough, duplicating the code to build the `HEALPix` C++ code could work.

I won't have time to work on this, anyone wants to try implementing this? @lpsinger @DanielLenz 

The latest HEALPix code is at:
https://github.com/healpy/healpixmirror/tree/trunk

lpsinger:
Fixes #576.

This should work but for the two issues mentioned in #576 that must first be fixed in the HEALPix svn repository.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -462,10 +462,17 @@ setup(
             },
         ),
         (
+            "sharp",
+            {
+                "pkg_config_name": "libsharp",
+                "local_source": "healpixsubmodule/src/common_libraries/libsharp",
+            },
+        ),
+        (
             "healpix_cxx",
             {
-                "pkg_config_name": "healpix_cxx >= 3.40.0",
-                "local_source": "healpixsubmodule/src/cxx/autotools",
+                "pkg_config_name": "healpix_cxx >= 3.60.0",
+                "local_source": "healpixsubmodule/src/cxx",
             },
         ),
     ],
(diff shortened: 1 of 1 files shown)


lpsinger:
@mreineck, the build is failing with this error message:

```
In file included from healpy/src/_sphtools.cpp:635:0:
1675/home/travis/build/healpy/healpy/build/temp.linux-x86_64-2.7/include/healpix_cxx/weight_utils.h:71:10: error: ‘unique_ptr’ in namespace ‘std’ does not name a template type
1676     std::unique_ptr<weight_utils_detail::FullWeightImpl> impl;
1677          ^
1678
```

Do we need to build with `-std=c++11`?

mreineck:
> Do we need to build with -std=c++11?

Yes, that will be necessary.

lpsinger:
@zonca, the build is now passing. This is ready for review.

zonca:
thanks @lpsinger for making this work!

PR #595, opened 2020-01-09
Support download of pixel weights at NSIDE 8192

zonca:
Pixel weights at NSIDE 8192 are too large to be added to
the `healpy-data` repo, had to attach them to a release.

See https://github.com/healpy/healpy-data/issues/1

### DIFF
--- a/healpy/sphtfunc.py
+++ b/healpy/sphtfunc.py
@@ -28,6 +28,7 @@ from scipy.integrate import trapz
 from astropy.utils import data
 
 DATAURL = "https://healpy.github.io/healpy-data/"
+DATAURL_MIRROR = "https://github.com/healpy/healpy-data/releases/download/"
 
 from . import _healpy_sph_transform_lib as sphtlib
 from . import _sphtools as _sphtools
@@ -213,8 +214,8 @@ def map2alm(
         if use_weights:
             raise RuntimeError("Either use pixel or ring weights")
         with data.conf.set_temp("dataurl", DATAURL), data.conf.set_temp(
-            "remote_timeout", 30
-        ):
+            "dataurl_mirror", DATAURL_MIRROR
+        ), data.conf.set_temp("remote_timeout", 30):
             pixel_weights_filename = data.get_pkg_data_filename(
                 "full_weights/healpix_full_weights_nside_%04d.fits" % nside,
                 package="healpy",


DanielLenz:
Thanks for working on that, looks good to me!

We should come up with a strategy for future nsides though, both on the computation and on the storage side of things.

mreineck:
> We should come up with a strategy for future nsides though, both on the computation and on the storage side of things.

I'm pretty convinced that the only valid strategy is trusting the user to compute their own weights (or download them by hand). Downloading 400MB files on demand (or actually opening any kind of network connection) without user confirmation is a pattern I definitely do not want to see in scientific software I'm using.

ntessore:
Would it be possible to make an optional pip package for `healpy-data` which contains these large files? As far as I am aware, the allowed size of a PyPI entry can be increased on a per-package basis.

Using pixel weights could then fail with a ModuleNotFoundError is the auxiliary files are not installed.

mreineck:
I know I'm trying to make my own stuff obsolete here, but maybe this is worth thinking about...

Map analysis using pixel weights is only usable for `lmax<=1.5*nside`, which is better than nothing,but far from what people are trying to achieve in cosmology (`lmax>=2*nside`), so in order to make use of this analysis mode, most pipelines would have to increase their `nside` by a factor of 2 (which makes them run more slowly in other places). I wonder if most people would not be best served by an approach like the one proposed in https://github.com/healpy/healpy/issues/676, which gives them fine control over the band limit and the desired accuracy, while at the same time being much more efficient than the "old" iterative map analysis.

ntessore:
@mreineck That sounds very interesting and I would be happy to try it out. But even if that is the best solution going forward, the pixel weights work well right now, so it would be nice to have a simple solution to get the weights installed where downloading is not an option, and manual installation is difficult.

The `lmax` thing is quite interesting, I wasn't aware. Should it then not throw an error if `lmax>1.5*nside` with `use_pixel_weights=True`?

```py3
>>> import numpy as np
>>> import healpy as hp
>>> nside = 64
>>> lmax = 2*nside
>>> m = np.random.randn(hp.nside2npix(nside))
>>> a = hp.map2alm(m, lmax=lmax, use_pixel_weights=True)
>>> np.shape(a)
(8385,)
```

mreineck:
> Should it then not throw an error if lmax>1.5*nside with `use_pixel_weights=True`?

Very good point; I think that would be quite helpful!
On the other hand ... if the analyzed map contains frequencies above that band limit, then the `map2alm` call will produce bad results no matter which `lmax` is specified. Harmonic analysis on the sphere is really a nasty topic...

Issue #592, opened 2019-12-04
Update CFITSIO bundled in healpy to last version


zonca:
it contains cfitsio 3.36 from Dec 2013, while HEASARC recommends using
at least version 3.44 from April 2018 for security reasons
(https://heasarc.gsfc.nasa.gov/fitsio/). I know that recent versions 
add an extra dependency on libcurl (which apparently is not universally pre-installed),
but is there a strong reason to keep shipping something with potential security issues ?

_Originally posted by @hivon in https://github.com/healpy/healpy/issues/590#issuecomment-561597207_

zonca:
fixed in #597

PR #597, opened 2020-01-14
update cfitsio to 3.47

zonca:
https://github.com/healpy/healpy/issues/592

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,6 @@
 Release in progress
 
+* Updated to cfitsio 3.48 (used only if missing) <https://github.com/healpy/healpy/pull/597>
 * Local datapath for pixel weights <https://github.com/healpy/healpy/pull/611>
 
 Release 1.13.0 3 Dec 2019, included in HEALPix 3.60


zonca:
@mreineck @lpsinger  I have trouble updating CFITSIO,
it seems like I can build libcfitsio, but the HEALPix configure script cannot find it.
See https://travis-ci.com/github/zonca/healpy/jobs/344641606#L1843

Would you have any suggestion on what is wrong?

mreineck:
`configure` complains because it can find the cfitsio library file, but it doesn't contain the `fngrwll` symbol. Most likely this happens because cfitsio was compiled without large file support, see, e.g., https://heasarc.gsfc.nasa.gov/docs/software/fitsio/c/f_user/node13.html
But I do not understand why this should have suddenly been turned off ...

mreineck:
I just checked the Travis build log ... large file support seems to be enabled. Weird.

mreineck:
Is there any way to access the `config.log` and other files generated by `configure` during the Travis build?

zonca:
@mreineck sure, I can reproduce it on my Ubuntu, it was just easier to upload everything to a throwaway repo:

https://github.com/zonca/build_logs_cfitsio

mreineck:
Thanks!
Here is what I think is happening: `configure` does actually find libcfitsio.so, but when it tries to link against it in order to look for the `ffgnrwll` symbol, `libcfitsio` itself requires `libcurl` and that is not found (for a reason I don't understand).
To be honest, I don't understand how `libcfitsio` is built during the configuration of `Healpix_cxx` if it is not found on the system. It seems as if `pkgconfig` is trying to do this and failing.
In any case I don't think that Healpix has been requiring a pre-installed `libcfitsio` for some time now; if the installation via `pkgconfig` worked up to now, that was a lucky accident.

zonca:
I think `libcurl` is there, but during configure:

```
configure:17569: x86_64-linux-gnu-gcc -pthread -o conftest -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O2 -Wall -g -      
fstack-protector-strong -Wformat -Werror=format-security -fopenmp -I/home/zonca/zonca/p/software/healpy/build/temp.linux-x86_64-3.7/ 
include  -fopenmp  conftest.c -lcfitsio   -L/home/zonca/zonca/p/software/healpy/build/temp.linux-x86_64-3.7/lib -lcfitsio -lm >&5
```

this should also have `-lcurl`

mreineck:
I agree, but this should be set automatically by the `pkg-config` description file coming with `cfitsio`. I don't know how this can be done from the "customer" side, because some versions of `libcfitsio` depend on curl and some don't.

mreineck:
I'm glad that this won't be an issue any more at some point in the future ;-)

mreineck:
I suspect that the `cfitsio` in question is installed without a `pkgconfig` file, and therefore `configure` falls back to the bare `-lcfitsio` flag (see lines 49-53 in `src/cxx/configure.ac`)

zonca:
`lib/pkgconfig/cfitsio.pc`

is

```
prefix=/home/zonca/zonca/p/software/healpy/build/temp.linux-x86_64-3.7
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: cfitsio
Description: FITS File Subroutine Library
URL: https://heasarc.gsfc.nasa.gov/fitsio/
Version: 3.48
Libs: -L${libdir} -lcfitsio
Libs.private: -lm 
Cflags: -I${includedir}
```

mreineck:
If `libcfitsio` needs the `-lcurl` flag when being linked to a binary, then it should be in there ...
I'm not absolutely sure that it is actually needed (since shared libraries can track their own dependencies to some extent), but if not, why does the `configure` test fail?
Can you try to add `-lcurl` temporarily to the "Libs:" line?

zonca:
we are using the static libraries, this worked on my machine!
thanks for your help debugging. let's hope for Travis to work as well.

mreineck:
OK, let's hope for the best!

zonca:
That was it, thanks!

Issue #609, opened 2020-04-16
astropy.utils.data download race condition from multiple processes


tskisner:
If multiple processes (for example in an mpi4py job) attempt to use astropy.utils.data to download and cache the same file, there exists a race condition where the processes can stomp on each other.  Here is a tarball containing a minimal working example that demonstrates the problem at NERSC.  Any computing center with more than one node and a shared home directory should see the same problem:

[astropy_cache.tar.gz](https://github.com/healpy/healpy/files/4489382/astropy_cache.tar.gz)

One possible solution would be:

1.  Provide a script with healpy that can pre-download all auxilliary data files to a central location

2.  Support using an environment variable or other mechanism to communicate where healpy should find these files.

## Running the Test

At NERSC, load any python stack that has a compatible mpi4py (either the NERSC
supported one or cmbenv).  For example:

    module load python
    pip install --upgrade --user astropy
    pip install --user healpy

OR

    module use /global/common/software/cmb/cori/default/modulefiles
    module load cmbenv
    source cmbenv

Next, ensure that the astropy download cache is empty, or move the
directory out of the way if you prefer:

    rm -rf ~/.astropy/cache/download

Now submit the example included in this tarball:

    sbatch run.slurm

## Test Results

The central problem is a race condition.  It is possible that by sheer luck one
process will download the file before the others try and no error will occur.  I
have purposely chosen the job concurrency to be high, so that the collision
between processes is more likely.  If your job happens to work, remove the
astropy download directory and try again.

Here is a typical error from one process:
```
Proc 40: Traceback (most recent call last):
Proc 40:   File "/global/u1/k/kisner/git/toast/examples/astropy_cache/./data_test.py", line 56, in <module>
    main()
Proc 40:   File "/global/u1/k/kisner/git/toast/examples/astropy_cache/./data_test.py", line 44, in main
    data = rot.rotate_map_alms(data, use_pixel_weights=True)
Proc 40:   File "/global/homes/k/kisner/.local/cori/3.7-anaconda-2019.10/lib/python3.7/site-packages/healpy/rotator.py", line 417, in rotate_map_alms
    m, use_pixel_weights=use_pixel_weights, lmax=lmax, mmax=mmax
Proc 40:   File "/global/homes/k/kisner/.local/cori/3.7-anaconda-2019.10/lib/python3.7/site-packages/healpy/sphtfunc.py", line 216, in map2alm
    package="healpy",
[...]

PR #611, opened 2020-05-05
Check datapath for pixel weights

zonca:
@tskisner, do you think this is enough to fix https://github.com/healpy/healpy/issues/609 ?

Please send feedback also on the docs.

### DIFF
--- a/healpy/sphtfunc.py
+++ b/healpy/sphtfunc.py
@@ -224,18 +224,29 @@ def map2alm(
     nside = pixelfunc.get_nside(maps)
     check_max_nside(nside)
 
+    pixel_weights_filename = None
     if use_pixel_weights:
         if use_weights:
             raise RuntimeError("Either use pixel or ring weights")
-        with data.conf.set_temp("dataurl", DATAURL), data.conf.set_temp(
-            "dataurl_mirror", DATAURL_MIRROR
-        ), data.conf.set_temp("remote_timeout", 30):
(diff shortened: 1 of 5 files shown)


healpy/sphtfunc.py
```diff
@@ -224,18 +230,29 @@ def map2alm(
...
-                "full_weights/healpix_full_weights_nside_%04d.fits" % nside,
-                package="healpy",
-            )
-    else:
-        pixel_weights_filename = None
+        filename = "full_weights/healpix_full_weights_nside_%04d.fits" % nside
+        if datapath is not None:
+            pixel_weights_filename = os.path.join(datapath, filename)
+            if os.path.exists(pixel_weights_filename):
+                warnings.warn(
```
tskisner, line 241: This could potentially generate many lines of output if `map2alm()` is called hundreds of times (e.g. by TOAST doing coordinate rotation once per detector).  However, we can filter out healpy warnings if needed, so that is fine.

tskisner:
@zonca, this PR adds the needed arguments to `map2alm()`, but there is no way to pass those arguments when using higher level code (like `Rotator` methods).  Any ideas?

zonca:
ok added to `rotate_map_alm`, anywhere else?

tskisner:
A quick "grep" of the source looks like it is only used within sphtfunc.py and rotator.py, and it looks like `datapath` is already an option for anafast and smoothing.  So I think you got all the functions, thanks!

PR #613, opened 2020-05-28
Setup conda on Readthedocs

zonca:


zonca:
[healpy fix readthedocs](https://trello.com/c/do7ocHfW/227-healpy-fix-readthedocs)

setup.py
```diff
@@ -76,79 +76,6 @@ def __init__(self, dist):
         build_clib.__init__(self, dist)
         self.build_args = {}
 
-    def autotools_path(self):
```
lpsinger, line 79: Eeek! Don't we need this stuff?

zonca: I think it was needed only for readthedocs

zonca: isn't it easier to ask people to install `automake` and `pkg-config` as requirements?

lpsinger: No, this wasn't needed for readthedocs, it was needed for building wheels.

zonca: ok, I'll revert that

lpsinger: Actually, never mind, I think this is OK. I misread the diff.

PR #617, opened 2020-06-13
Add support for line integral convolution (LIC)

mpetroff:
This pull request adds line integral convolution support based on the `alice3` executable from Healpix 3.60.

Since `alice3` is structured as a standalone executable, I copied the C++ code and modified it such that it could be called as a function from Cython, although refactoring the code upstream to support both uses would have been another possibility.

The other main change made is to add an option to the visualization functions to allow the current axes to be reused. When combined with a partially transparent colormap, this allows the LIC output to be plotted over another map.

I also fixed a compilation bug I came across in `_sphtools.pyx`, which I believe is related to the default language level change in Cython 3.0.

Here's an example of the new functionality:
```python3
import numpy as np
import matplotlib.pyplot as plt
from matplotlib.cm import get_cmap
from matplotlib.colors import ListedColormap
import healpy as hp

cmap_colors = get_cmap('binary', 256)(np.linspace(0, 1, 256))
cmap_colors[..., 3] = 0.4  # Make colormap partially transparent
cmap = ListedColormap(cmap_colors)

m = hp.read_map('wmap_band_iqumap_r9_9yr_K_v5.fits', (0, 1, 2), verbose=False)
I, Q, U = hp.smoothing(m, np.deg2rad(5))
lic = hp.line_integral_convolution(Q, U)

lic = hp.smoothing(lic, np.deg2rad(0.5))

hp.mollview(np.log(1 + np.sqrt(Q**2 + U**2) * 100), cmap='inferno', cbar=False)
hp.mollview(lic, cmap=cmap, cbar=False, reuse_axes=True, title='WMAP K')

plt.savefig('wmapk.png', dpi=150)
```

wmapk

### DIFF
--- a/healpy/src/_line_integral_convolution.pyx
+++ b/healpy/src/_line_integral_convolution.pyx
@@ -1,0 +1,124 @@
+import numpy as np
+cimport numpy as np
+cimport cython
+from _common cimport RING, Healpix_Map, ndarray2map
+from pixelfunc import npix2nside
+from healpy.sphtfunc import alm2map, Alm
+
+cdef extern from "_line_integral_convolution_lib.cc":
+    cdef void lic_main(const Healpix_Map[double] &Q, const Healpix_Map[double] &U,
+                       const Healpix_Map[double] &th, Healpix_Map[double] &hit,
+                       Healpix_Map[double] &tex, Healpix_Map[double] &mag,
+                       int steps, int kernel_steps, double step_radian,
+                       double polmin, double polmax)
+
+
+def line_integral_convolution(
+    Q,
+    U,
+    steps=100,
+    kernel_steps=50,
+    step_radian=0.003,
+    pol_min=-1e30,
+    pol_max=1e30,
+    rand_seed=42,
+    ell=None,
+    texture=None,
+    modulate=False,
+):
+    """Computes line integral convolution of polarized map.
+
+    Parameters
+    ----------
+    Q : array-like, shape (Npix,)
+      Input Stokes Q map. Must be in RING ordering.
+    U : array-like, shape (Npix,)
+      Input Stokes U map. Must be in RING ordering.
+    steps : int
+      Number of steps to use for each line of the convolution (default: 100).
+    kernel_steps : int
+      Extent of the convolution kernel (in steps, <= steps) (default: 50).
+    step_radian : float
+      Size (in radians) of each step in the convolution (default: 0.003).
+    pol_min : float
+      Minimum value for the polarization magnitude (default: -1e30).
+    pol_max : float
+      Maximum value for the polarization magnitude (default: 1e30).
+    rand_seed : int
+      Seed for the random number generator; only used when texture is not
+      provided (default: 42).
+    ell : int or None
+      The ell value at which the generated background texture has power
+      (default: ``None``). If ``ell < 0``, it is set to ``ell=2*nside``. If
+      ``None``, the generated background texture will contain white noise.
+    texture : array-like, shape (Npix,) or None
+      Background texture map (default: ``None``). Must be in RING ordering. If
+      ``None``, a background texture will be generated based on the values of
+      ``rand_seed`` and ``ell``.
+    modulate : bool
+      Whether or not to modulate the convolved texture with the polarization
+      intensity (default: ``False``).
+
+    Returns
+    -------
+    lic_texture : Line integral convolved texture.
+
+    Example
+    -------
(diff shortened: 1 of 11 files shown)


zonca:
impressive! thank you very much,
I will have time to check this out in a couple of weeks!

zonca:
@mreineck are you ok with copying this line integral code into `healpy`? or are you willing to backport the changes necessary to call from Python into HEALPix C++?

healpy/src/_sphtools.pyx
```diff
@@ -574,6 +574,7 @@ def rotate_alm(alm not None, double psi=0, double theta=0, double phi=0, matrix=
         del AI, AG, AC
 
 
+@cython.cdivision(True)
```
zonca, line 577: why do you add this?

healpy/src/_line_integral_convolution.pyx
```diff
@@ -0,0 +1,124 @@
...
+    step_radian=0.003,
+    pol_min=-1e30,
+    pol_max=1e30,
+    rand_seed=42,
+    ell=None,
+    texture=None,
+    modulate=False,
+):
+    """Computes line integral convolution of polarized map.
+
```
zonca, line 30: can you please explain a bit what is a line integral convolution?

healpy/src/_line_integral_convolution.pyx
```diff
@@ -0,0 +1,124 @@
...
+      Extent of the convolution kernel (in steps, <= steps) (default: 50).
+    step_radian : float
+      Size (in radians) of each step in the convolution (default: 0.003).
+    pol_min : float
+      Minimum value for the polarization magnitude (default: -1e30).
+    pol_max : float
+      Maximum value for the polarization magnitude (default: 1e30).
+    rand_seed : int
+      Seed for the random number generator; only used when texture is not
+      provided (default: 42).
```
zonca, line 49: what about using `None` as default seed?

healpy/src/_line_integral_convolution.pyx
zonca: and the most important parameters like `texture`

healpy/src/_line_integral_convolution.pyx
mreineck: C++ codes does not have the concept of a `None` seed meaning "initialize the RNG in a nonreproducible way".  And I'd argue that this is actually the better choice in a scientific context. In game deveopment, things would probably be different :)

mreineck:
> @mreineck are you ok with copying this line integral code into healpy? or are you willing to backport the changes necessary to call from Python into HEALPix C++?

I'd love to avoid the code duplication, but I simply don't have the time at the moment to make the necessary adjustments. Fortunately it's not a huge amount of code, so it will probably not become a maintenance burden.

healpy/src/_sphtools.pyx
mpetroff: I was getting a compiler error with Cython 3.0, which I think is related to the default language level change in Cython 3.0.

healpy/src/_line_integral_convolution.pyx
mpetroff: I moved the PRNG use into the Cython file, so the C++ restrictions don't apply (of course `None` could just have been interpreted to seed the C++ PRNG with a random number from NumPy's PRNG). The fixed seed was used to ensure reproducible output; the default seed value was copied from the `alice3` executable.

healpy/src/_line_integral_convolution.pyx
mpetroff: I added a bit more information to the doc string.

mpetroff:
I made an attempt at patching HEALPix C++ so the code could be shared:
[alice3-diff.txt](https://github.com/healpy/healpy/files/4961528/alice3-diff.txt)

The only change in behavior to the existing `alice3` executable is to use double-precision instead of single-precision (the alternative would be to make a variant of Healpy's `ndarray2map` function for single-precision), but I've only tested it to the extent of making sure it doesn't crash.

mreineck:
Wow, thanks a lot!
I will definitely integrate this as soon as I have some time.
But I don't think it should hold up the imminent next releases of Healpix/healpy.

zonca:
@mreineck just a reminder that it would be useful to have this patch into HEALPix C++

mreineck:
Thanks for the reminder! I had to tweak the patch a bit to preserve the original output images, but this should work now.

PR #624, opened 2020-07-09
Fix const correctness in Numpy ufunc API

lpsinger:
Fixes https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=964655.

### DIFF
--- a/healpy/src/_healpy_pixel_lib.cc
+++ b/healpy/src/_healpy_pixel_lib.cc
@@ -39,7 +39,7 @@
    ang2pix
 */
 template<Healpix_Ordering_Scheme scheme>static void
-  ufunc_ang2pix(char **args, npy_intp *dimensions, npy_intp *steps, void *func)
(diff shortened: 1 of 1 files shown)


mreineck:
I like it!

lpsinger:
The tests are failing because the const qualifier was aded in numpy/numpy#15355. The patch here needs to be a bit more subtle. Either we can use error and warning pragmas to temporarily adjust the compiler error settings in the affected code, or we can add a macro/typedef that depends on the Numpy version.

mreineck:
Do I understand this correctly: the `numpy` developers changed the interface in a way that breaks existing code (OK so far), but now all the broken code needs a really ugly hack to be compatible with old and new `numpy` versions? Yuck ...

lpsinger:
Correct.

lpsinger:
Adding explicit casts to `PyUFuncGenericFunction` fixed it. Builds are passing now, please review.

Issue #619, opened 2020-06-15
Make default verbose=False?


adrn:
It looks like the default behavior for functions (e.g., `healpy.smoothing()`) is `verbose=True`. Should the default be changed to `verbose=False`? I frequently use `healpy` from an IPython notebook, and the output is often clogged by verbose output, especially when looping over maps. Just raising this for discussion - I don't feel very strongly about this!

zonca:
In that case less experienced users would never see the messages.

My proposal would be to switch all functions to use `warnings` (`smoothing` is not),
and then explain in the docs how to do:

    warnings.filterwarnings(action='ignore', module='healpy' )

to suppress them.
Any better suggestion? @lpsinger?

lpsinger:
I agree with @adrn. Functions in Python modules should be silent by default.

zonca:
no-one would ever set all verbose to True, so they would be useless, their purpose is to help the novice users when running interactively.
I think my proposal is good enough, it allows more advanced uses to just stop all `healpy` verboseness with just 1 line of code. I implemented also a small function to make it easier.

```
In [5]: hp.smoothing(np.arange(12), fwhm=1)                                                                                          
/home/zonca/zonca/p/software/healpy/healpy/sphtfunc.py:823: UserWarning: Sigma is 1459.876638 arcmin (0.424661 rad) 
  sigma * 60 * 180 / np.pi, sigma
/home/zonca/zonca/p/software/healpy/healpy/sphtfunc.py:828: UserWarning: -> fwhm is 3437.746771 arcmin
  sigma * 60 * 180 / np.pi * (2.0 * np.sqrt(2.0 * np.log(2.0)))
/home/zonca/zonca/p/software/healpy/healpy/sphtfunc.py:823: UserWarning: Sigma is 0.000000 arcmin (0.000000 rad) 
  sigma * 60 * 180 / np.pi, sigma
/home/zonca/zonca/p/software/healpy/healpy/sphtfunc.py:828: UserWarning: -> fwhm is 0.000000 arcmin
  sigma * 60 * 180 / np.pi * (2.0 * np.sqrt(2.0 * np.log(2.0)))
Out[5]: 
array([0.97237589, 1.96986319, 2.68512681, 2.84879047, 4.85208202,
       4.59814393, 5.64311657, 7.05682575, 7.65126875, 8.64875605,
       9.36401967, 9.52768333])

In [6]: hp.disable_warnings()                                                                                                        

[...]

PR #630, opened 2020-07-22
Rework verbose

zonca:
See https://github.com/healpy/healpy/issues/619#issuecomment-662212313


adrn:
I definitely like having the ability to disable the messages globally! 

The only thought I have (which might be filed under "bikeshedding") is that these messages seem like they are debugging log messages rather than warnings, so it feels a little wrong to be using the `warnings` machinery for them vs. a Python `logging.Logger` instance. Another option would be to implement a logger, set the default log level to `logging.DEBUG`, and make it available so that users can do, e.g.,:
```python
from healpy.log import logger
logger.setLevel(logging.WARN)
```

zonca:
@adrn I agree `logging` would be more suitable, but I used it in the past and found it confusing, I like the simplicity of `warnings` just having on/off and no multiple loggers.

PR #631, opened 2020-07-22
Fix unclosed files 

zonca:
`pytest` was complaining (`ResourceWarning: unclosed file`) files were not closed by `read_map`, fixed that. 

Also, hit this `astropy` issue https://github.com/astropy/astropy/issues/10594, put a workaround converting `pathlib.Path` to `str`

### DIFF
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -42,6 +42,9 @@ standard_column_names = {
     6: ["II", "IQ", "IU", "QQ", "QU", "UU"],
 }
 
+allowed_paths = tuple(six.string_types)
+if sys.version >= "3.4":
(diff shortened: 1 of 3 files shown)


zonca:
@mpetroff would you have 5 min to double-check this?

mpetroff:
The now duplicate allowed path types code (lines 684-686) in the `_get_hdu` function should be removed.

Also, as `pathlib.PosixPath` is an subclass of `pathlib.Path`, it doesn't need to be added to the list of allowed path types, but it doesn't hurt either:
```python3
>>> isinstance(pathlib.PosixPath('test'), pathlib.Path)
True
```
Otherwise, it looks fine.

Issue #623, opened 2020-07-07
Release of healpy 1.14.0


zonca:
I'd like to start preparing the release of healpy 1.14.0, jointly with a HEALPix release.
@hivon, @mreineck are you fine with that? what will be the HEALPix version tag?

can I get the current trunk from SVN as the HEALPix C++ that will released with the next version of HEALPix?

mreineck:
No changes in Healpix C++ since the last healpy release, as far as I see :)

hivon:
@mreineck : there was an edit in libsharp automake files to set the library version number, which changes the libsharp.so file name.
And how about the rotmatrix.h and pointing.h name conflicts discussed in https://sourceforge.net/p/healpix/bugs/104/ ?
Will we go with the suggested path specification for Heasoft or rather change the path to those *.h files ?

@zonca : the next Healpix release is 3.70 and should come out by July 17 (or July 24 at worst) or otherwise wait until late August.

mreineck:
Ah, thanks for pointing this out ... somehow my `svn log` didn't show up anything relevant and I had forgotten about that change already! (It's getting hard to remember how svn works; I'm only using it a few times per year)

Concerning the header files: I'd leave them as they are for the moment.  I'm currently consolidating all my C++ code into a single directory structure (from which sub-packages can be extracted at need), and there I'm using a directory scheme and namespaces to avoid this kind of collision in the future. If you are interested, have a look at https://gitlab.mpcdf.mpg.de/mtr/ducc.
But this is still in an early stage, and the focus is currently on providing Python interfaces for the relevant functionality. I don't even have a working `autotools` machinery for it yet.

zonca:
ok, thanks @hivon, not sure I'll make it by July,

anyway, sometime in July or August, I'll update the C++ sources with current trunk, finish up some of the open issues, and release 1.14.0

hivon:
Hi @zonca,

do you have a clearer idea of when healpy 1.14 will be ready to ship out ?
As for Healpix it is either by July 26 (this Friday), or after Aug 24.
It would be better to avoid Healpix and healpy releases just 1 or 2 
weeks apart.

Cheers,
Eric

zonca:
@hivon I'll let you know tomorrow

zonca:
@hivon I'll release `healpy` 1.14.0 either today or tomorrow, then you can package it up into HEALPix on Friday

hivon:
Hi @zonca,

thanks !
[...]

PR #632, opened 2020-07-22
updated C++ sources to 3.70

zonca:
See https://github.com/healpy/healpy/issues/623


lpsinger:
I am looking to update the Debian and MacPorts packages, but I don't see HEALPix 3.70 in https://sourceforge.net/projects/healpix/files/.

zonca:
it is going to be released on Friday see #623

lpsinger:
It would make life a little bit easier if Healpix_cxx was released _before_ Healpy because Healpix_cxx is a dependency of Healpy. However, I do know that there is the chicken-and-egg problem due to the monolithic packaging of HEALPix.

Issue #639, opened 2020-09-16
Use alice3 from HEALPix C++


zonca:
@mreineck included the `alice3` patches from #617 to support line integral convolution into HEALPix C++,
and I have updated the https://github.com/healpy/healpixmirror to the last version of HEALPix from SVN.

**NB**: `healpy` master is still pointing to the last released HEALPix and will change that only when HEALPix is ready for the next release.

It would be useful to have a Pull Request that:
* updates the `healpixsubmodule` to the tip of `healpixmirror`
* removes the C++ files added in https://github.com/healpy/healpy/pull/617
* properly imports directly from HEALPix C++

@mpetroff would you like to take care of this? no hurry, I'll need it in a couple of months.

mpetroff:
Yes, I can look at this. I already have a partial patch in https://github.com/healpy/healpy/pull/617#issuecomment-662594550.

PR #640, opened 2020-09-18
Use HEALPix C++ copy of alice3 for LIC

mpetroff:
Update to use HEALPix C++ copy of `alice3` for LIC, closing #639.

I'm not particularly familiar with the build system, so there might be a better way of configuring `setup.py` for this.


zonca:
excellent, thank you!
I will keep this open and merge it just before the next joint release of HEALPix and `healpy`.
We do this to avoid people that have both HEALPix and `healpy` installed to have different versions of the C++ library.

PR #654, opened 2021-01-22
Add requests to tests_require list

spenczar:
`test_pixelweights.py` imports `requests`, so it should be in `setup.py`'s list of requirements: https://github.com/healpy/healpy/blob/master/healpy/test/test_pixelweights.py#L3

Before this, I was unable to run `python setup.py test`. It works after this change.


zonca:
thanks!

PR #657, opened 2021-02-05
Replace np.bool with np.bool_ and np.float with np.float64

erykoff:
The np.bool and np.float types have been deprecated in numpy 1.20 and
the correct versions have been available for a very long time.


zonca:
thanks!

PR #656, opened 2021-02-05
Add GitHub Actions workflow to build wheels

lpsinger:

PR #658, opened 2021-02-08
Drop support for Python 2.7-3.5, which are past end of life

lpsinger:


lpsinger:
Note, this will make it easier to build wheels (#656).

zonca:
thanks!

PR #659, opened 2021-02-08
Replace obsolete ez_setup.py with PEP 518 pyproject.toml

lpsinger:
This simplifies and modernizes bootstrapping setuptools and including build dependencies.


mreineck:
This is almost certainly off-topic, but I found it while reading the new `pyproject.toml`...

Why is there still a dependency on `six`? `healpy` is now Python 3 only, correct?

lpsinger:
> This is almost certainly off-topic, but I found it while reading the new `pyproject.toml`...
> 
> Why is there still a dependency on `six`? `healpy` is now Python 3 only, correct?

No, not yet. That's in a separate PR. See #658.

lpsinger:
Note, this will make it easier to build wheels (#656).

zonca (approved):
it looks good to me, go ahead and merge when ready

PR #660, opened 2021-02-08
Fix macOS Catalina build

lpsinger:
Fixes #18016. Inspired by numpy/numpy#18016.


lpsinger:
Note, I haven't actually tested this because I do not have access to a macOS 11 machine. But it *should* fix it.

Issue #649, opened 2020-12-14
Drop pytest-runner requirement as it is deprecated


badshah400:
Thanks for the great library.

I would like to suggest the dropping of pytest-runner from [setup.py](https://github.com/healpy/healpy/blob/bb73236067e3af713a6750226603179d01c6301c/setup.py#L483) due to the following:

> pytest-runner depends on deprecated features of setuptools and relies on features that break security mechanisms in pip. For example ‘setup_requires’ and ‘tests_require’ bypass pip --require-hashes. See also https://github.com/pypa/setuptools/issues/1684.
>
> It is recommended that you:
>
> * Remove ‘pytest-runner’ from your ‘setup_requires’, preferably removing the setup_requires option.
> * Remove ‘pytest’ and any other testing requirements from ‘tests_require’, preferably removing the setup_requires option.
> * Select a tool to bootstrap and then run tests such as tox

(from <https://pypi.org/project/pytest-runner/>)

zonca:
it would be great if you can provide a pull request to fix this. thanks!

badshah400:
I saw your PR #650 and thought that was it. My patch would have been the same. Any reason why that doesn't work?

zonca:
Cannot find pytest, need to add that as requirement on travis

PR #662, opened 2021-02-08
remove pytest-runner

zonca:
See #649


zonca:
this is a rabbit hole of awkward building errors. Giving up.

PR #665, opened 2021-02-15
Do not assume that OBJECT is a string

lpsinger:
LIGO/Virgo probability maps sometimes set OBJECT to an integer to indicate an event ID:

https://git.ligo.org/lscsoft/ligo.skymap/-/blob/v0.5.0/CHANGES.rst#L11-13


lpsinger:
@zonca, could we please have a release to pick up this bug fix? That will also exercise the new wheels built on GitHub Actions.

zonca:
yes, will do later this week

PR #674, opened 2021-03-27
Replace travis tests with GitHub Actions

### LINKED ISSUE #670: Migrate from travis to Github actions
Travis now charges open-source, so we need to migrate to Github actions.
Some specific help is available at https://docs.github.com/en/actions/learn-github-actions/migrating-from-travis-ci-to-github-actions

@lpsinger do you have time to take care of this? otherwise I can work on it.

I would like to have this setup before doing next release.

lpsinger:
Use cibuildwheel to test the actual binary packages.

Fixes #670.


zonca:
Thanks!
Would it be useful to also add 1 configuration testing external cfitsio and healpix? Maybe Linux only

zonca (approved):

zonca:
thanks!

lpsinger:
> Would it be useful to also add 1 configuration testing external cfitsio and healpix? Maybe Linux only

It sure would. That should run tests against the built source distribution. I'm a bit too busy to do that myself at the moment.

zonca:
I think it is quite a corner case by now, maybe @tskisner builds HEALPix separately from healpy at NERSC?
I think it would be useful to do a test now that we changed a bit of the building setup.
@tskisner would you be able to test if the `main` branch of `healpy` builds fine with an external HEALPix? We need the trunk version of healpix (or see the submodule in healpy)

tskisner:
I no longer attempt to build healpix separately from healpy.  However, I do override the compilers and flags:

https://github.com/hpc4cmb/cmbenv/blob/6d1b02895eaba73f4fdbaf236d9c9a2318076124/pkgs/healpy.sh#L25

This allows us to use (for example) the Intel compilers when building the underlying healpix compiled library and extensions.

PR #678, opened 2021-04-02
Suggestion of a new nest parameter in healpy.smoothing() 

MaudeCharmetant:
I suggest implementing a new "nest" parameter in healpy.sphtfunc.smoothing() so that when the input map is in NESTED ordering, the function takes care of reordering it to RING for the smoothing and back to NESTED before the output. I have tried to follow the way "nest" argument where used in other Healpy functions, but maybe the description in the docstring can be improved and maybe the ordering back to NESTED can be removed and you could just specify that healpy.sphtfunc.smoothing() will return a RING ordered map no matter the input. 

I hope this helps a bit or at least serves as a base to improve the clarity of the smoothing function.

### DIFF
--- a/healpy/sphtfunc.py
+++ b/healpy/sphtfunc.py
@@ -945,10 +945,10 @@ def smoothing(
     verbose : bool, optional
       If True prints diagnostic information. Default: True
     nest : bool, optional
-      If True, ordering scheme is NESTED. Default: False (RING)
-      The map ordering is modified by this function, the input map array should
-      be in RING ordering, if not this function will convert it to RING ordering 
-      and order the output back to NESTED. 
+      If True, the input map ordering is assumed to be NESTED. Default: False (RING)
+      This function will temporary reorder the NESTED map into RING to perform the
+      smoothing and order the output back to NESTED. If the map is in RING ordering 
+      no internal reordering will be performed. 
 
     Returns
     -------
@@ -1029,10 +1029,7 @@ def smoothing(
     output_map[masks] = UNSEEN
     
     if nest:
-        ordering = "NESTED"
         output_map = pixelfunc.reorder(output_map, inp=None, out=None, r2n=True, n2r=False)
-    else:
-        ordering = "RING" 
 
     return output_map
 
--- a/healpy/sphtfunc.py
+++ b/healpy/sphtfunc.py
@@ -972,10 +972,7 @@ def smoothing(
     check_max_nside(nside)
     
     if nest:
-        ordering = "NESTED"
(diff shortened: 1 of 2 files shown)


healpy/sphtfunc.py
```diff
@@ -943,6 +944,11 @@ def smoothing(
       If given, the directory where to find the weights data.
     verbose : bool, optional
       If True prints diagnostic information. Default: True
+    nest : bool, optional
+      If True, ordering scheme is NESTED. Default: False (RING)
```
zonca, line 948: ordering scheme "of the input map is assumed to be NESTED"

healpy/sphtfunc.py
```diff
@@ -943,6 +944,11 @@ def smoothing(
...
+      The map ordering is modified by this function, the input map array should
```
zonca, line 949: `reorder` makes copy of the array, so we are not modifying the input.
however I agree we need to mention that internally we are running `reorder`

healpy/sphtfunc.py
```diff
@@ -1015,6 +1027,12 @@ def smoothing(
             output_map.append(alm2map(alm, nside, pixwin=False, verbose=verbose))
         output_map = np.array(output_map)
     output_map[masks] = UNSEEN
+    
+    if nest:
+        ordering = "NESTED"
+        output_map = pixelfunc.reorder(output_map, inp=None, out=None, r2n=True, n2r=False)
+    else:
+        ordering = "RING"
```
zonca, line 1035: we probably dont need to redefine `ordering` here.

zonca (commented):
very good, thanks for your help with this! I added feedback for minor improvements

zonca:
about reordering the output, let's keep that, we don't want by default the function to have an output ordering different than the input ordering.

can you also add a line to `CHANGELOG.rst` mentioning this?

MaudeCharmetant:
Thank you for all of your comments! I have implemented them, let me know if we need further changes, especially on the phrasing of things. I have updated CHANGELOG.rst but I think I was supposed to update it in my branch (which I did afterward) and not submit a new pull request, sorry.

healpy/sphtfunc.py
MaudeCharmetant: Yes true, thank you! I have removed the "ordering" and also the else condition which is not essential I suppose.

zonca (approved):

zonca:
thanks for your contribution, much appreciated!

MaudeCharmetant:
Thank you for Healpy, this is such a great tool!

Issue #675, opened 2021-03-29
healpy enabling warnings


kadrlica:
This is a continuation of discussions about the default verbosity (#619) and the implementation of the `warnings` module to control verbosity (#630).  

I've found that [healpy.enable_warnings](https://github.com/healpy/healpy/blob/8b853d89062f0f704231dbc2770ea29028beacbc/healpy/__init__.py#L121) resets other UserWarning filters outside the healpy module. Looking at the code, this isn't surprising:

https://github.com/healpy/healpy/blob/8b853d89062f0f704231dbc2770ea29028beacbc/healpy/__init__.py#L121-L122

A simple example:
```
import warnings
import healpy as hp

warnings.simplefilter("ignore", category=UserWarning)
warnings.warn("Warning 1") # this is ignored

hp.enable_warnings()
warnings.warn("Warning 2") # this is not ignored
```

I don't think this is the desired behavior. I think that `hp.enable_warnings` should be the inverse of `hp.disable_warnings` and should act only on warnings from the `healpy` module.

zonca:
Sure, would you be able to provide a pull request fixing this?

lpsinger:
IMHO, Healpy rather overdoes it with warnings. I think in many cases, what are presented as warnings should in fact be log messages presented with the `logging` module, and hidden by default (eg. debug level). I subscribe to the opinion that libraries should be silent by default.

kadrlica:
I was not going to get into this, but... I completely agree with @lpsinger and @adrn about both the verbosity and (mis)use of `warnings`. I think that the `logging` module is the way to go. Novice users (the undergrads I work with for example) freak out when they get blasted with warnings...

kadrlica:
Regarding the PR, I'm not an expert on warning control, but would it be enough to just invert the `disable_warnings` call? This implementation seems to work on my simple test case (at least as a monkey patch). However, I'm not sure if it is turning on other warnings that we don't want to turn on. Any idea what tests we'd want to implement?

```
def enable_warnings():
    warnings.filterwarnings(action="always", module="healpy")
```

zonca:
see #679

PR #679, opened 2021-04-03
warnings to logging and changing defaults for write_map, read_map

zonca:
Following #675 and previous issues, I switched the library to using `logging`, most of the previous warnings became debug or info messages which are not shown by default, just a minor important ones became `logging.warning`.

The most important objective of warnings was to warn users of the future change of the default setting for `hp.write_map` and `hp.read_map`.
Until the last release (1.14.0), whatever precision a map is in memory, it is written to disk in single precision (`float32`) and when read, it is converted to double precision (`float64`). As was written in the warning, now instead `healpy` will maintain whatever precision the map is in both when reading and writing.
I am a bit worried that a user upgrading is not notified about this, it is in the docstring and the changelog but there is no other notification, any suggestions here?

The other warning that was present was in `synalm` and `synfast` about the order of  `C_ell`, going for `new=False` to `new=True`, basically changing from TT TE EE.... to TT EE BB. This seems too dangerous for old code. I prefer to remove this notice and leave the default as it is.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -23,6 +23,11 @@ is very welcome to propose new features.
 * transform maps to Spherical Harmonics space and back using multi-threaded C++ routines
 * compute Auto and Cross Power Spectra from maps and create map realizations from spectra
 
+Changelog
+---------
+
+Review the changes in each release in the `CHANGELOG on Github <https://github.com/healpy/healpy/blob/main/CHANGELOG.rst>`_.
+
 Citing
 ------
 
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -23,6 +23,28 @@ is very welcome to propose new features.
 * transform maps to Spherical Harmonics space and back using multi-threaded C++ routines
 * compute Auto and Cross Power Spectra from maps and create map realizations from spectra
 
+Verbosity
+---------
+
+Starting from 1.15.0, `healpy` uses the `logging` module instead of `warnings`.
+By default `healpy` will only print warnings and errors, to configure logging,
+you can use :py:func:configure_logging, which by default enables all logging as
+in `healpy` 1.14.0.
+For more advanced configuration you can access the "healpy" logger with::
+
+    import logging
+    log = logging.getLogger("healpy")
+
+For more details see the `Python documentation <https://docs.python.org/3/library/logging.html>`_.
+
+All the `verbose` keywords (except :py:func:remove_dipole) are now deprecated and
+will be removed in the future.
+You can disable deprecation warnings with::
+
+    import warnings
+    from astropy.utils.exceptions import AstropyDeprecationWarning
+    warnings.simplefilter('ignore', category=AstropyDeprecationWarning)
+
 Changelog
 ---------
 
--- a/healpy/__init__.py
+++ b/healpy/__init__.py
@@ -104,10 +104,3 @@ from .fitsfunc import write_map, read_map, read_alm, write_alm, write_cl, read_c
 from ._masktools import dist2holes_healpy as dist2holes
 from ._hotspots import hotspots_healpy as hotspots
 from ._line_integral_convolution import line_integral_convolution
-
-# Default null handler prevents printing all log messages if the user doesn't
-# configure `logging`
-# See https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library
-import logging
-
-logging.getLogger("healpy").addHandler(logging.NullHandler())
--- a/healpy/__init__.py
+++ b/healpy/__init__.py
@@ -21,8 +21,6 @@
 compute spherical harmonics tranforms on them.
 """
 
-import warnings
-
 from .version import __version__
 
 from .pixelfunc import (
@@ -105,18 +103,10 @@ from .zoomtool import mollzoom, set_g_clim
 from .fitsfunc import write_map, read_map, read_alm, write_alm, write_cl, read_cl
 from ._masktools import dist2holes_healpy as dist2holes
 from ._hotspots import hotspots_healpy as hotspots
-from ._line_integral_convolution import line_integral_convolution
 
+# Default null handler prevents printing all log messages if the user doesn't
+# configure `logging`
+# See https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library
+import logging
 
-def disable_warnings():
-    """Disable all healpy warnings messages for the current session
-
-    Warnings from individual functions can be disabled setting
-    ``verbose=False``.
-    Warnings can be re-enabled calling ``hp.enable_warnings()``.
-    """
-    warnings.filterwarnings(action="ignore", module="healpy")
-
-
-def enable_warnings():
-    warnings.simplefilter("always")
+logging.getLogger("healpy").addHandler(logging.NullHandler())
--- a/healpy/__init__.py
+++ b/healpy/__init__.py
@@ -103,6 +103,7 @@ from .zoomtool import mollzoom, set_g_clim
 from .fitsfunc import write_map, read_map, read_alm, write_alm, write_cl, read_cl
 from ._masktools import dist2holes_healpy as dist2holes
 from ._hotspots import hotspots_healpy as hotspots
+from ._line_integral_convolution import line_integral_convolution
 
 # Default null handler prevents printing all log messages if the user doesn't
 # configure `logging`
--- a/healpy/__init__.py
+++ b/healpy/__init__.py
@@ -21,6 +21,8 @@
 compute spherical harmonics tranforms on them.
 """
 
+import logging
+
 from .version import __version__
 
 from .pixelfunc import (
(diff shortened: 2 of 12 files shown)


healpy/fitsfunc.py
```diff
@@ -190,14 +184,8 @@ def write_map(
...
-        warnings.warn(
-            "The default dtype of write_map() will change in a future version: "
-            "explicitly set the dtype if it is important to you",
-            category=FutureWarning,
-        )
-        dtype = [np.float32 for x in m]
-        # Change this at some point to:
-        # dtype = [x.dtype for x in m]
+        dtype = [x.dtype for x in m]
+        logging.info("setting the output map dtype to %s", str(dtype))
```
lpsinger, line 188:
```suggestion
        logging.info("setting the output map dtype to %s", dtype)
```

healpy/fitsfunc.py
```diff
@@ -135,7 +124,12 @@ def write_map(
     extra_header=(),
     overwrite=False,
 ):
-    """Writes a healpix map into a healpix file.
+    """Writes a healpix map into a healpix FITS file.
+
+    WARNING: starting from healpy 1.15.0, if you do not specify `dtype`,
+    the map will be written to disk with the same precision it is stored in memory.
+    Previously, by default `healpy` wrote maps in `float32`.
+    To reproduce the same behaviour of `healpy` 1.14.0 and below, set `dtype=np.float32`.
```
lpsinger, line 132:
```suggestion
    .. warning:: Starting from healpy 1.15.0, if you do not specify `dtype`,
       the map will be written to disk with the same precision it is stored in memory.
       Previously, by default `healpy` wrote maps in `float32`.
       To reproduce the same behaviour of `healpy` 1.14.0 and below, set `dtype=np.float32`.
```

Also change similar text from other docstrings to ReStructuredText admonitions so that they are given emphasis.

healpy/__init__.py
```diff
@@ -107,16 +105,9 @@
...
-    ``verbose=False``.
-    Warnings can be re-enabled calling ``hp.enable_warnings()``.
-    """
-    warnings.filterwarnings(action="ignore", module="healpy")
-
-
-def enable_warnings():
-    warnings.simplefilter("always")
\ No newline at end of file
+logging.getLogger("healpy").addHandler(logging.NullHandler())
```
lpsinger, line 113: It would be a good idea to see if there is any benefit to directly tapping into Astropy's warning system.

healpy/fitsfunc.py
```diff
@@ -347,9 +338,8 @@ def read_map(
       the header number to look at (start at 0)
     h : bool, optional
       If True, return also the header. Default: False.
-    verbose : bool, optional
-      If True, print a number of diagnostic messages, call hp.disable_warnings() to
-      disable warnings for all functions.
+    verbose : bool, not supported anymore
```
lpsinger, line 341: Deprecate this so that it can be removed in a future release. Astropy's utils module has some features for deprecating arguments.

healpy/sphtfunc.py
```diff
@@ -1242,8 +1200,7 @@ def beam2bl(beam, theta, lmax):
 
     nx = len(theta)
     nb = len(beam)
-    if nb != nx:
-        warnings.warn("beam and theta must have same size!")
+    assert nb == nx, "beam and theta must have same size!"
```
lpsinger, line 1203: Seems like a ValueError would be more appropriate here than an assertion.

healpy/test/test_visufunc.py
```diff
@@ -93,7 +93,7 @@ def test_azeqview_ma_nocrash(self):
     def test_cmap_colors(self):
         # Get a built-in colormap
         name = 'viridis'
-        cmap = matplotlib.cm.get_cmap(name)
+        cmap = copy.copy(matplotlib.cm.get_cmap(name))
```
lpsinger, line 96: Not sure why this is part of this changeset.

zonca:
@kadrlica would you please review this PR? suggestions very welcome

lpsinger (changes requested):
Instead of using the global logging helper functions, please get a logger object and use it.

```python
log = logging.getLogger("healpy")
...
log.warning(...)
```

healpy/projaxes.py
```diff
@@ -915,7 +891,7 @@ def create_colormap(cmap, badcolor="gray", bgcolor="white"):
             "newcm", cmap0._segmentdata, cmap0.N
         )
     else:
-        newcm = copy.deepcopy(cmap0)
+        newcm = copy.copy(cmap0)
```
zonca, line 894: @kadrlica compared to https://github.com/healpy/healpy/pull/627, I was getting a warning with `deepcopy` and not with `copy`

lpsinger: So this should probably go in a separate PR.

zonca: right, moved to https://github.com/healpy/healpy/pull/681

healpy/fitsfunc.py
zonca: see ca0473f863a3d299e0e9b4d07fffa3e8df547218

healpy/fitsfunc.py
zonca: this would generate warnings. isn't it acceptable to ask users to go through their code and remove all `verbose` arguments? it seems a quick fix that can be just done once (and will need to be done in the future).

lpsinger: A deprecation warning is least obtrusive way to tell users to do that.

lpsinger: https://docs.astropy.org/en/stable/api/astropy.utils.decorators.deprecated_renamed_argument.html#astropy.utils.decorators.deprecated_renamed_argument

zonca: ok

doc/index.rst
```diff
@@ -23,6 +23,33 @@ is very welcome to propose new features.
...
+
+    import warnings
+    from astropy.utils.exceptions import AstropyDeprecationWarning
+    warnings.simplefilter('ignore', category=AstropyDeprecationWarning)
+
+Changelog
+---------
+
+Review the changes in each release in the `CHANGELOG on Github <https://github.com/healpy/healpy/blob/main/CHANGELOG.rst>`_.
+
```
zonca, line 52: @lpsinger ok, I deprecated all verbose, added docs here about verbosity

healpy/__init__.py
```diff
@@ -107,16 +107,54 @@
...
+    log.setLevel(level)
+
+    # create console handler
+    handler = logging.StreamHandler()
+
+    # create formatter
+    formatter = logging.Formatter(log_format)
+    handler.setFormatter(formatter)
+
+    log.addHandler(handler)
```
zonca, line 160: @lpsinger it is not straightforward to configure `logging`, setting only the level is not enough you also need to add a handler, so better provide a function.

healpy/__init__.py
zonca: I think `logging` only except for deprecation is fine.

doc/index.rst
```diff
@@ -23,6 +23,33 @@ is very welcome to propose new features.
...
+For more details see the `Python documentation <https://docs.python.org/3/library/logging.html>`_.
+
+All the `verbose` keywords (except :py:func:remove_dipole) are now deprecated and
+will be removed in the future.
+You can disable deprecation warnings with::
+
+    import warnings
+    from astropy.utils.exceptions import AstropyDeprecationWarning
+    warnings.simplefilter('ignore', category=AstropyDeprecationWarning)
```
zonca, line 46: @lpsinger ok, I deprecated all verbose, added docs here about verbosity

healpy/__init__.py
zonca, line 160: @lpsinger it is not straightforward to configure logging, setting only the level is not enough you also need to add a handler, so better provide a function.

healpy/pixelfunc.py
```diff
@@ -1578,7 +1574,7 @@ def remove_dipole(
 
         lon, lat = R.vec2dir(dipole, lonlat=True)
         amp = np.sqrt((dipole * dipole).sum())
-        warnings.warn(
+        print(
```
kadrlica, line 1577: Should this also be a logging statement?

healpy/projaxes.py
```diff
@@ -46,9 +47,8 @@ class SphericalProjAxes(matplotlib.axes.Axes):
     coordprec : number of digit after floating point for coordinates display.
     format : format string for value display.
 
-    Notes
-    -----
-    Other keywords from Axes (see Axes).
+    .. note::
+       Other keywords from Axes (see Axes).
```
lpsinger, line 51: More incidental format changes that should go in a separate PR or changeset.

healpy/projaxes.py
```diff
@@ -194,8 +189,9 @@ def projmap(
         if vmin == vmax:
             vmin -= 1.0
             vmax += 1.0
-        cm, nn = get_color_table(vmin, vmax, img[w], cmap=cmap, norm=norm,
-                                 badcolor=badcolor, bgcolor=bgcolor)
+        cm, nn = get_color_table(
+            vmin, vmax, img[w], cmap=cmap, norm=norm, badcolor=badcolor, bgcolor=bgcolor
+        )
```
lpsinger, line 194: Incidental code style change, should go in a separate changeset.

healpy/projaxes.py
```diff
@@ -205,8 +201,6 @@ def projmap(
             norm=nn,
             interpolation="nearest",
             origin="lower",
-            vmin=vmin,
-            vmax=vmax,
```
lpsinger, line 209: Another unrelated change...

healpy/projaxes.py
```diff
@@ -293,9 +286,11 @@ def projplot(self, *args, **kwds):
...
                     linestyle, marker, color = matplotlib.axes._process_plot_format(fmt)
                 except:  # matplotlib 1.4 and later
-                    linestyle, marker, color = matplotlib.axes._axes._process_plot_format(
-                        fmt
-                    )
+                    (
+                        linestyle,
+                        marker,
+                        color,
+                    ) = matplotlib.axes._axes._process_plot_format(fmt)
```
lpsinger, line 293: Another unrelated change...

healpy/projaxes.py
```diff
@@ -536,16 +527,8 @@ def graticule(
...
-            warnings.warn(
-                "{0} {1} {2} {3}".format(
-                    pmin / dtor, pmax / dtor, mmin / dtor, mmax / dtor
-                )
-            )
         if not kwds.pop("force", False):
-            dpar, dmer = self._get_interv_graticule(
-                pmin, pmax, dpar, mmin, mmax, dmer, verbose=verbose
-            )
+            dpar, dmer = self._get_interv_graticule(pmin, pmax, dpar, mmin, mmax, dmer)
```
lpsinger, line 531: Another unrelated change...

healpy/projaxes.py
```diff
@@ -612,19 +595,18 @@ def graticule(
...
+        """Delete all graticules previously created on the Axes."""
         if hasattr(self, "_graticules"):
             for dum1, dum2, g in self._graticules:
                 for gl in g:
                     for l in gl:
                         if l in self.lines:
                             self.lines.remove(l)
                         else:
-                            warnings.warn("line not in lines")
+                            log.warning("line not in lines")
```
lpsinger, line 606: Is this going to spew a large amount of log chatter, since it's inside a triply nested loop?

healpy/rotator.py
```diff
@@ -55,18 +58,6 @@
     .value
 )
 
-
-class ConsistencyWarning(Warning):
-    """Warns for a problem in the consistency of data
-    """
-
-    pass
```
lpsinger, line 63: Needs a deprecation warning prior to removal.

zonca:
splitting into multiple PRs

PR #684, opened 2021-04-26
Added alpha channel in visualization

aasensio:
Transparency is still now allowed in Healpy. I've added a new `alpha` keyword to all map visualization routines to allow for this. To avoid returning errors when using colorbars with transparency (the error is produced by Matplotlib `imshow`), the colorbar is extracted from the original map without transparency.

### DIFF
--- a/healpy/projaxes.py
+++ b/healpy/projaxes.py
@@ -174,7 +174,7 @@ class SphericalProjAxes(matplotlib.axes.Axes):
         coord : {'G', 'E', 'C', None}
           The coordinate system of the map ('G','E' or 'C'), rotate
           the map if different from the axes coord syst.
-        map : array-like
+        alpha : array-like
           The alpha (transparency) map.
 
         Notes
--- a/healpy/projaxes.py
+++ b/healpy/projaxes.py
@@ -23,6 +23,7 @@ from . import rotator as R
 from . import pixelfunc
 import matplotlib
 import matplotlib.axes
+import matplotlib.pyplot as plt
 import numpy as np
 import copy
 
@@ -146,6 +147,7 @@ class SphericalProjAxes(matplotlib.axes.Axes):
         norm=None,
         rot=None,
         coord=None,
+        alpha=None,
         **kwds
     ):
         """Project a map on the SphericalProjAxes.
@@ -172,6 +174,8 @@ class SphericalProjAxes(matplotlib.axes.Axes):
         coord : {'G', 'E', 'C', None}
           The coordinate system of the map ('G','E' or 'C'), rotate
           the map if different from the axes coord syst.
+        map : array-like
+          The alpha (transparency) map.
 
         Notes
         -----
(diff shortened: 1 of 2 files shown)


healpy/projaxes.py
```diff
@@ -172,13 +174,21 @@ def projmap(
         coord : {'G', 'E', 'C', None}
           The coordinate system of the map ('G','E' or 'C'), rotate
           the map if different from the axes coord syst.
+        map : array-like
```
zonca, line 177: shouldn't this be `alpha`?

zonca (changes requested):
very good!
can you provide a simple example of using this?
I can then add that to the tutorials

healpy/projaxes.py
aasensio: Yes. Updated.

aasensio:
Here you have an example of use:

```
import numpy as np
import healpy as hp

NSIDE = 32
NPIX = hp.nside2npix(NSIDE)

m = np.arange(NPIX)
alpha = np.ones(NPIX)

vec = hp.ang2vec(np.pi / 2, np.pi / 4)
ipix_disc = hp.query_disc(nside=NSIDE, vec=vec, radius=np.radians(30))
alpha[ipix_disc] = 0.0

alpha = hp.sphtfunc.smoothing(alpha, sigma=0.2, verbose=False)

hp.mollview(m, alpha=alpha)
```

The code and output can be seen [here](https://gist.github.com/aasensio/459429fc53e1b30db179ad12e3ffae76)

zonca:
thanks @aasensio, what is a typical use you do of this functionality? I think it would be useful to have a notebook with a more concrete example so that users have a clearer idea of what this can be used for.

zonca (approved):

aasensio:
Great!
I implemented it because I'm playing with showing uncertainties in spherical maps. It is not clear at all to me that this is the proper way of showing uncertainty but there might be other use cases. I updated a trivial example [here](https://gist.github.com/aasensio/b7e5a93ce5de7b9a137527be6dff17c1) based on your tutorial for masking the CMB temperature maps. I simply show how the masked CMB map can be showed, also displaying an unrealistic smoothed mask. It makes little sense but it can give an idea of the applications.

healpy/visufunc.py
```diff
@@ -1329,14 +1360,16 @@ def azeqview(
             badcolor=badcolor,
             bgcolor=bgcolor,
             norm=norm,
+            alpha=alpha,
         )
         if cbar:
             im = ax.get_images()[0]
             b = im.norm.inverse(np.linspace(0, 1, im.cmap.N + 1))
             v = np.linspace(im.norm.vmin, im.norm.vmax, im.cmap.N)
+            mappable = plt.cm.ScalarMappable(norm=matplotlib.colors.Normalize(vmin=im.norm.vmin, vmax=im.norm.vmax), cmap=cmap)
```
zonca, line 1369: why do you need to build the `ScalarMappable` object? how is it related to the alpha channel?

zonca: @aasensio can you please explain why you need `ScalarMappable`?

aasensio: Sorry, had a busy couple of weeks. Colorbars in `plt.imshow` cannot be used when the alpha channel per pixel is different,  because the colormap is not properly defined. This `ScalarMappable` defines the range of colors of the image to build the colorbar as if there were no alpha channel and it is a workaround this issue with `plt.imshow`.

zonca:
follows in #696, otherwise I cannot rebase.

PR #685, opened 2021-04-30
Added convenice functions order2npix and npix2order.

DouglasBoubert:
I added two convenience functions that map from resolution order to number of pixels and back.

The only thing that these functions do is call nside2npix(order2nside(order)) and nside2order(npix2nside(npix)).

### DIFF
--- a/healpy/pixelfunc.py
+++ b/healpy/pixelfunc.py
@@ -131,6 +131,8 @@ __all__ = [
     "npix2nside",
     "nside2order",
     "order2nside",
+    "order2npix",
+    "npix2order",
     "nside2resol",
     "nside2pixarea",
     "isnsideok",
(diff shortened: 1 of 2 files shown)


healpy/pixelfunc.py
```diff
@@ -131,6 +131,8 @@
     "npix2nside",
     "nside2order",
     "order2nside",
+    "order2npix",
+    "npix2order",
```
zonca, line 135: can you also import them into `__init__.py`?

zonca (changes requested):

zonca:
Thanks for your contribution!

healpy/pixelfunc.py
DouglasBoubert: Ah, sorry for that! I've imported them.

zonca (approved):

PR #686, opened 2021-05-07
fix: import from Cython last

zonca:
See
https://stackoverflow.com/questions/21594925/error-each-element-of-ext-modules-option-must-be-an-extension-instance-or-2-t

Error trying to build the package:

```
error: each element of 'ext_modules' option must be an Extension
instance or 2-tuple
```

Noticed on:

* Linux
* Python 3.8
* Cython 0.29.22
* setuptools 51.1.2 

@hivon also noticed on Mac OS, Python 3.9.

However, it doesn't cause any error in the Github Actions build.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,6 @@ import errno
 import fnmatch
 import sys
 import shlex
-from Cython.Distutils import build_ext, Extension
 from distutils.sysconfig import get_config_var, get_config_vars
 import pkg_resources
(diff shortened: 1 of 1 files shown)


zonca:
There is another way of fixing this error, see https://github.com/healpy/healpy/pull/690

@lpsinger do you have a recommendation on how to fix this?

zonca:
#690 seems more reasonable to me, instead of changing order we are importing Extension from a different package.

PR #687, opened 2021-05-07
Remove note that we will change order in synalm and synfast

zonca:
I think we shouldn't change this, better keep supporting both ways keeping the same default we have now.


zonca:
this is only a change in the documentation

PR #688, opened 2021-05-07
Change default for read_map and write_map

zonca:
Until the last release (1.14.0), whatever precision a map is in memory, it is written to disk in single precision (`float32`) and when read, it is converted to double precision (`float64`). As was written in the warning, now instead `healpy` will maintain whatever precision the map is in both when reading and writing.
I am a bit worried that a user upgrading is not notified about this, it is in the docstring and the changelog but there is no other notification, any suggestions here?

### DIFF
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -765,3 +765,9 @@ def getformat(t):
             return conv[np.dtype(t[0].type)]
     except:
         pass
+
+    raise ValueError(
+        "healpy could not understand the equivalent FITS datatype of {}, please open an issue on the healpy Github repository".format(
+            str(t)
+        )
+    )
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -716,6 +716,8 @@ def getformat(t):
         np.dtype(np.complex64): "C",
         np.dtype(np.complex128): "M",
     }
+    if hasattr(t, "type"):
+        t = t.type
     try:
         if t in conv:
             return conv[t]
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -719,8 +719,6 @@ def getformat(t):
         np.dtype(np.complex64): "C",
         np.dtype(np.complex128): "M",
     }
-    if hasattr(t, "type"):
-        t = t.type
     try:
         if t in conv:
             return conv[t]
@@ -755,5 +753,15 @@ def getformat(t):
         if type(t[0]) is str:
             l = max(len(s) for s in t)
             return "A%d" % (l)
+    except:
+        pass
+    try:
+        if np.dtype(t.type) in conv:
+            return conv[np.dtype(t.type)]
+    except:
+        pass
+    try:
+        if np.dtype(t[0].type) in conv:
+            return conv[np.dtype(t[0].type)]
     except:
         pass
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -130,7 +130,8 @@ def write_map(
 ):
     """Writes a healpix map into a healpix FITS file.
 
-    WARNING: starting from healpy 1.15.0, if you do not specify `dtype`,
+    .. warning::
+    Starting from healpy 1.15.0, if you do not specify `dtype`,
     the map will be written to disk with the same precision it is stored in memory.
     Previously, by default `healpy` wrote maps in `float32`.
     To reproduce the same behaviour of `healpy` 1.14.0 and below, set `dtype=np.float32`.
@@ -173,8 +174,9 @@ def write_map(
       internally from the numpy datatype to the fits convention. If a list,
       the length must correspond to the number of map arrays.
       Default: use the data type of the input array(s)
-      (WARNING: this changed in 1.15.0, previous versions saved in float32
-      by default)
+      .. note::
+      this changed in 1.15.0, previous versions saved in float32
+      by default
     overwrite : bool, optional
       If True, existing file is silently overwritten. Otherwise trying to write
       an existing file raises an OSError (IOError for Python 2).
@@ -303,7 +305,8 @@ def read_map(
     """Read a healpix map from a fits file.  Partial-sky files,
     if properly identified, are expanded to full size and filled with UNSEEN.
 
-    WARNING: starting from healpy 1.15.0, if you do not specify `dtype`,
+    .. warning::
+    Starting from healpy 1.15.0, if you do not specify `dtype`,
     the map will be read in memory with the same precision it is stored on disk.
     Previously, by default `healpy` wrote maps in `float32` and then upcast to
     `float64` when reading to memory. To reproduce the same behaviour of `healpy`
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -328,7 +328,7 @@ def read_map(
       types for each field. In that case, the length of the list must
       correspond to the length of the field parameter.
       If None, keep the dtype of the input FITS file
-      Default: None
+      Default: Preserve the data types in the file
     nest : bool, optional
       If True return the map in NEST ordering, otherwise in RING ordering;
       use fits keyword ORDERING to decide whether conversion is needed or not
--- a/healpy/fitsfunc.py
+++ b/healpy/fitsfunc.py
@@ -128,7 +128,12 @@ def write_map(
     extra_header=(),
     overwrite=False,
 ):
-    """Writes a healpix map into a healpix file.
+    """Writes a healpix map into a healpix FITS file.
+
+    WARNING: starting from healpy 1.15.0, if you do not specify `dtype`,
+    the map will be written to disk with the same precision it is stored in memory.
+    Previously, by default `healpy` wrote maps in `float32`.
+    To reproduce the same behaviour of `healpy` 1.14.0 and below, set `dtype=np.float32`.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


healpy/fitsfunc.py
```diff
@@ -721,6 +721,8 @@ def getformat(t):
         np.dtype(np.complex64): "C",
         np.dtype(np.complex128): "M",
     }
+    if hasattr(t, "type"):
+        t = t.type
```
zonca, line 725: @lpsinger this doesn't come out in the unit tests.
But if I run the tutorial notebook, I have:

```
wmap_map_I.dtype
dtype('>f4')
```

If I try to write it to disk:

```
> hp.write_map("my_map.fits", wmap_map_I, overwrite=True)
/extrahome/software/healpy/healpy/fitsfunc.py:199: UserWarning: setting the output map dtype to [dtype('>f4')]
  warnings.warn("setting the output map dtype to %s" % str(dtype))

---------------------------------------------------------------------------
VerifyError                               Traceback (most recent call last)
<ipython-input-15-38a2a4f140b4> in <module>
----> 1 hp.write_map("my_map.fits", wmap_map_I, overwrite=True)

/extrahome/software/healpy/healpy/fitsfunc.py in write_map(filename, m, nest, dtype, fits_IDL, coord, partial, column_names, column_units, extra_header, overwrite)
    241             mm2 = np.asarray(mm)
    242             cols.append(
--> 243                 pf.Column(
    244                     name=cn,
    245                     format="1024%s" % curr_fitsformat,

~/miniconda3/envs/healpyclean/lib/python3.8/site-packages/astropy/io/fits/column.py in __init__(self, name, format, unit, null, bscale, bzero, disp, start, dim, array, ascii, coord_type, coord_unit, coord_ref_point, coord_ref_value, coord_inc, time_ref_pos)
    619                 msg.append(indent(val[1]))
    620 
--> 621             raise VerifyError('\n'.join(msg))
    622 
    623         for attr in KEYWORD_ATTRIBUTES:

VerifyError: The following keyword arguments to Column were invalid:
    Column format option (TFORMn) failed verification: Illegal format `1024None`. The invalid value will be ignored for the purpose of formatting the data in this column.
```

I am adding a unit test about this.

zonca: It seems like it is caused by endianness, I don't understand why, if the input array is `>f4`, this patch is needed, if it is `f4`, it is not necessary. See test https://github.com/healpy/healpy/pull/688/commits/540ef2ddf90eab20f6c58f70fddf00f6604bbeb3

mreineck:
> I am a bit worried that a user upgrading is not notified about this, it is in the docstring and the changelog but there is no other notification, any suggestions here?

Sorry for the late reply. If we want to ensure that the users make an informed decision, the only way is to remove the defaults and have them select the data type in the form of, say, "single", "double", or "as_in_file". If we don't do that (and I completely understand if you don't want to do it), some users will not notice the change.

zonca:
* code that is specifying single (`np.float32`) or double (`np.float64`) or "same as input" (`None`) is not impacted, whatever the version.

## `healpy` 1.14.0

code that is not specifying a dtype will write maps in float32 and convert them to float64 on loading

## `healpy` 1.15.0 with new defaults

code that is not specifying a dtype will keep the same dtype of the input 

##  `healpy` 1.15.0 with no defaults

code that is not specifying a dtype will get an exception and the user will need to explicitly choose one of the 3 options

I actually think removing defaults is probably the best option here. @lpsinger @keskitalo @tskisner @giuspugl what do you think?

keskitalo:
Python code should not try to guess what the user intends to do. In this application, I would interpret that as using the data type in file, unless the user explicitly asked for something else. That works for boolean masks, hit maps and floating point. The only complication arises when reading multiple columns with different data types, in which case usual upcasting rules should be observed.

Forcing users to select a data type would serve in the transition period but would lead to unnecessary arguments when sensible defaults are available. It would also break existing code.

My vote is on `dtype=None` by default, translating into keeping the original data type on file or in the provided map array.

giuspugl:
Yeah agree with what @keskitalo  proposes. I'd avoid breaking existing code

mreineck:
> Yeah agree with what @keskitalo proposes. I'd avoid breaking existing code

In my eyes the problem is that existing code may be "broken" already now without the uers actually realizing it. One of the things that prompted this discussion was  https://github.com/healpy/healpy/issues/585#issuecomment-553763380.
It's quite possible that more latent bugs of that kind are lurking out there...

keskitalo:
An excellent example of a Python code "guessing" what the user would want to do. And with detrimental results. I understand that during the transition it helps to have all users check their save/load data types. Perhaps make specifying the save/load data type mandatory for now but add dropping it to the road map?

zonca:
@lpsinger would you have a recommendation for the `write_map`/`read_map` defaults?

lpsinger:
I can look at this on the evening of May 27.

zonca:
@lpsinger ok, then it doesn't need to be the 27th, if you can do before June 2nd it would be great, thanks.

healpy/fitsfunc.py
```diff
@@ -722,6 +716,8 @@ def getformat(t):
         np.dtype(np.complex64): "C",
         np.dtype(np.complex128): "M",
     }
+    if hasattr(t, "type"):
```
lpsinger, line 719: This does indeed seem to correctly handle types of specified, non-native endianness. However, this is a very low-level way to go about it. The `astropy.io.fits` module internally has more robust code to map dtypes to FITS TTYPEn cards. I think it would be advisable to rewrite much of this so that it uses higher-level `astropy.io.fits` interfaces and doesn't muck about with dtypes and FITS headers directly.

If `t` is a dtype, then I think that it will always have the `type` attribute. However, if `t` is a string like `>f4`, then the function won't correctly handle it; you would need to first call `np.dtype(t)`.

This function could certainly be simplified to reduce the number of different conversion attempts.

lpsinger (changes requested):

healpy/fitsfunc.py
```diff
@@ -128,7 +128,12 @@ def write_map(
     extra_header=(),
     overwrite=False,
 ):
-    """Writes a healpix map into a healpix file.
+    """Writes a healpix map into a healpix FITS file.
+
+    WARNING: starting from healpy 1.15.0, if you do not specify `dtype`,
+    the map will be written to disk with the same precision it is stored in memory.
+    Previously, by default `healpy` wrote maps in `float32`.
+    To reproduce the same behaviour of `healpy` 1.14.0 and below, set `dtype=np.float32`.
```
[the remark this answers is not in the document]
zonca: was planning to do the styling in following pull request, but as you have noted them on this PR, will do here.

healpy/fitsfunc.py
```diff
@@ -183,14 +188,8 @@ def write_map(
...
-        warnings.warn(
-            "The default dtype of write_map() will change in a future version: "
-            "explicitly set the dtype if it is important to you",
-            category=FutureWarning,
-        )
-        dtype = [np.float32 for x in m]
-        # Change this at some point to:
-        # dtype = [x.dtype for x in m]
+        dtype = [x.dtype for x in m]
+        warnings.warn("setting the output map dtype to %s" % str(dtype))
```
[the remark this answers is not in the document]
zonca: it is in #693

healpy/fitsfunc.py
```diff
@@ -304,6 +303,12 @@ def read_map(
     """Read a healpix map from a fits file.  Partial-sky files,
     if properly identified, are expanded to full size and filled with UNSEEN.
 
+    WARNING: starting from healpy 1.15.0, if you do not specify `dtype`,
```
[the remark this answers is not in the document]
zonca: will do WARNING for 1.15.0 to make it more visible, in the future can move into Notes (also notes should be after parameters and returns)

healpy/fitsfunc.py
zonca: it seems `fits.Column` objects still need to manually provide the format.
do you mean to use `Table` objects? https://docs.astropy.org/en/stable/io/unified.html#table-io-fits

zonca: I agree, it is a large change, targeting next release #699

healpy/test/test_fitsfunc.py
```diff
@@ -311,5 +311,10 @@ def test_getformat():
     assert getformat(["DD", "CCC"]) == "A3"
 
 
+def test_writemap_newdefault(tmp_path):
+    m = np.zeros(12, dtype=">f4")
+    write_map(tmp_path / "test_map.fits", m, overwrite=True)
+
+
```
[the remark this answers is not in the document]
zonca: rewrote the unit test

healpy/fitsfunc.py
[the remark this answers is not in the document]
lpsinger: OK, is the idea here that this is such a significant change that for a time you want there to be a warning _after_ the deprecated behavior has been removed?

healpy/test/test_fitsfunc.py
```diff
@@ -311,5 +311,32 @@ def test_getformat():
...
+        with pf.open(filename) as f:
+            assert f[1].data.field(0).dtype.name == dtype
+            np.testing.assert_array_almost_equal(f[1].data.field(0), m)
+
+
+def test_writemap_typestr(tmp_path):
+    filename = tmp_path / "test_map.fits"
+    m = np.zeros(12, dtype=">f4")
+    write_map(filename, m, overwrite=True)
+    assert pf.open(filename)[1].data.field(0).dtype.name == "float32"
```
lpsinger, line 338: This doesn't test that endianness is handled correctly. To do that, you need to write an asymmetric byte sequence: `np.ones(12)` would do the trick. Then you need to check that the bytes of the file written to disk are identical regardless of the endianness of the Numpy array.

lpsinger (changes requested):

healpy/test/test_fitsfunc.py
zonca: ok, done, it is more important that what I read back is the correct array, so I am testing that.

zonca:
rebased on `main`

healpy/fitsfunc.py
```diff
@@ -758,3 +755,13 @@ def getformat(t):
...
+    try:
+        if np.dtype(t.type) in conv:
+            return conv[np.dtype(t.type)]
+    except:
+        pass
+    try:
+        if np.dtype(t[0].type) in conv:
+            return conv[np.dtype(t[0].type)]
+    except:
+        pass
```
zonca, line 767: @lpsinger this should have less unexpected side-effects than the previous fix

PR #696, opened 2021-05-25
Support for alpha channel in visualization

zonca:
Following from #684

- Added alpha channel in visualization
- Updated docstring with the alpha channel
- fix: alpha is already None
- format with black


aasensio:
Fantastic!

PR #697, opened 2021-05-26
remove pytest-runner from pyproject.toml

zonca:


lpsinger:
You absolutely still need pytest-runner if you want `python setup.py test` to work.

zonca:
what if we deprecate `python setup.py test`?

like it's done in the astropy package template (changing the error message):

https://github.com/astropy/package-template/blob/bf8ce319e8ec99c38ab80c8af8c0437c16de0474/%7B%7B%20cookiecutter.package_name%20%7D%7D/setup.py#L18-L41

lpsinger:
Fine with me, but that would involve more updates to setup.py, updates to the CI script, and updates to the instructions too.

zonca:
The CI script already uses pytest

PR #698, opened 2021-06-04
Replace all deprecated numpy aliases with builtins

duncanmmacleod:
This PR replaces all instances of the [deprecated numpy aliases](https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated) with the appropriate (identical) builtin. This should have absolutely zero functional impact, but will remove a bunch of `DeprecationWarning`s for users of `numpy >=1.20.0`.


zonca:
Thanks!

PR #708, opened 2021-06-11
reimplement spin 0 transforms

ntessore:
CC @zonca @mreineck 

This flips the signs for spin 0 transforms `alm2map_spin` and `map2alm_spin`.

Note that there was an existing `test_spin0` that checked the other sign convention specifically.

### DIFF
--- a/healpy/src/_sphtools.pyx
+++ b/healpy/src/_sphtools.pyx
@@ -88,16 +88,14 @@ def map2alm_spin_healpy(maps, spin, lmax = None, mmax = None):
 
     Warnings
     --------
-    This function previously made an exception for ``spin=0`` transforms, where
-    it flipped the sign of the output such that :math:`a_{lm}^+` matches the
-    output :math:`a_{lm}` of `map2alm`.  The new behaviour follows the HEALPix
-    convention, in which :math:`a_{lm}^+ = -a_{lm}` for a ``spin=0`` transform.
+    Previously, ``spin=0`` was an invalid parameter value, and silently
+    returned undefined results.
 
     """
     spin = int(spin)
 
     if spin == 0:
-        return [-map2alm(mm, niter=3, lmax=lmax, mmax=mmax) for mm in maps]
+        return [-map2alm(mm, niter=0, lmax=lmax, mmax=mmax) for mm in maps]
 
     maps_c = [np.ascontiguousarray(m, dtype=np.float64) for m in maps]
 
@@ -179,10 +177,8 @@ def alm2map_spin_healpy(alms, nside, spin, lmax, mmax=None):
 
     Warnings
     --------
-    This function previously made an exception for ``spin=0`` transforms, where
-    it flipped the sign of the input such that :math:`a_{lm}^+` matches the
-    input :math:`a_{lm}` of `alm2map`.  The new behaviour follows the HEALPix
-    convention, in which :math:`a_{lm}^+ = -a_{lm}` for a ``spin=0`` transform.
+    Previously, ``spin=0`` was an invalid parameter value, and silently
+    returned undefined results.
 
     """
     spin = int(spin)
--- a/healpy/src/_sphtools.pyx
+++ b/healpy/src/_sphtools.pyx
@@ -15,6 +15,8 @@ from libcpp cimport bool as cbool
 
 from _common cimport tsize, arr, xcomplex, Healpix_Ordering_Scheme, RING, NEST, Healpix_Map, Alm, ndarray2map, ndarray2alm, rotmatrix
 
+from . import _healpy_sph_transform_lib as sphtlib
+
 cdef double UNSEEN = -1.6375e30
 cdef double rtol_UNSEEN = 1.e-7 * 1.6375e30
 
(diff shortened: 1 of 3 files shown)


mreineck (approved):

healpy/src/_sphtools.pyx
```diff
@@ -137,6 +137,11 @@ def map2alm_spin_healpy(maps, spin, lmax = None, mmax = None):
         if mask is not False:
             m[mask] = UNSEEN
 
+    # flip sign for spin 0 to match HEALPix convention
+    if spin == 0:
+        for alm in alms:
+            alm *= -1
+
```
zonca, line 144: can you please explain what was the behaviour before and after this in the docstring? so if a user notice something wrong they can understand what happened.

zonca (changes requested):
can you also add a line to `CHANGELOG.rst` about this?

ntessore:
Oh my, I forgot to git add the changelog to to the commit. I'll also try and understand the test failure, which I'm not seeing locally on macOS, so I'll try on another Ubuntu machine.

ntessore:
I've removed the second part of the test which was failing under Linux. I don't understand what's going on there yet, but that's outside the scope of this PR either way.

There's two other potential follow-ups:

- Improve the docstrings of `map2alm_spin` and `alm2map_spin`.
- Add support for weights in `map2alm_spin`.

zonca:
I think improvement of docstring would be great

hivon:
Hi @zonca @mreineck
I am not sure to understand why the patch is made in the python code, instead of directly in the underlying sharp library
which broke the sign convention to begin with. Now Healpix has to apply the same ugly patch.
This assumes that libsharp will keep that undocumented "feature" (supposed to be inaccessible) for ever. 
The whole point of doing joint releases of Healpix, healpy and libsharp (with only a handful of developers) is to apply patches at the right place, instead of multiplying release-dependent patches and counter-patches in various places.

zonca:
I agree with @hivon,
@mreineck if we can fix this upstream either in `libsharp` or HEALPix it is definitely better.

mreineck:
> I am not sure to understand why the patch is made in the python code, instead of directly in the underlying sharp library
which broke the sign convention to begin with.

That's a valid point.

However, I just re-checked the `libsharp` sources, and it turns out that if you run any of the spin-weighted transforms with a spin of 0, the code will completely ignore the second map/a_lm. It was simply not built for this application scenario, and I cannot fox this at short notice.

I'm afraid the only possible way forward for the imminent release is to raise an error when a spin-weighted tansform is called with a spin of 0.

@ntessore when you called this in the past, did you ever have anything different from 0 in the second map/a_lm of the pair?

ntessore:
@mreineck I have used the `_spin` functions in a generic spin `s` context before, but I suppose I never checked the correctness for `spin=0`. Because now that I was doing it again, the issue came up immediately.

I will change this PR to raise a `ValueError` for nonpositive spin.

ntessore:
@mreineck I have reimplemented the spin-0 transforms in terms of the unspinned `map2alm` and `alm2map` functions. Could you make sure I'm using these correctly? (I'd be grateful even if you don't want to merge this PR.)

healpy/src/_sphtools.pyx
```diff
@@ -83,7 +85,20 @@ def map2alm_spin_healpy(maps, spin, lmax = None, mmax = None):
     -------
     alms : list of 2 arrays
       list of 2 alms
+
+    Warnings
+    --------
+    This function previously made an exception for ``spin=0`` transforms, where
```
mreineck, line 91: Maybe just be blunt and say "This function previously returned undefined results when called with `spin=0`? Same for the alm2map direction.

healpy/src/_sphtools.pyx
```diff
@@ -83,7 +85,20 @@ def map2alm_spin_healpy(maps, spin, lmax = None, mmax = None):
...
+    it flipped the sign of the output such that :math:`a_{lm}^+` matches the
+    output :math:`a_{lm}` of `map2alm`.  The new behaviour follows the HEALPix
+    convention, in which :math:`a_{lm}^+ = -a_{lm}` for a ``spin=0`` transform.
+
     """
+    spin = int(spin)
+
+    if spin == 0:
+        return [-map2alm(mm, niter=3, lmax=lmax, mmax=mmax) for mm in maps]
```
mreineck, line 100: I'm not sure about the `niter=3` call here; as far as I understand, iterative analysis is handled at a higher level and therefore I would specify `niter=0` here. If the toplevel `map2alm` was called with some non-zero `niter`, control flow will come past this location several times anyway.

mreineck:
Just to have it said explicitly: I'm perfectly happy with these patches and think they should go in.

healpy/src/_sphtools.pyx
```diff
@@ -83,7 +85,18 @@ def map2alm_spin_healpy(maps, spin, lmax = None, mmax = None):
     -------
     alms : list of 2 arrays
       list of 2 alms
+
+    Warnings
```
zonca, line 89: can you please use the `warnings` directive? https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html#warnings

zonca:
> Just to have it said explicitly: I'm perfectly happy with these patches and think they should go in.

ok, if the fix is not doable in `libsharp` let's do it in `healpy`.
It is better than having a broken function.

healpy/src/_sphtools.pyx
ntessore: Since you’re using numpydoc, that’s effectively what I have done. See the docs. Unless you want me to specifically use the directive for some reason?

PR #711, opened 2021-06-18
accept `None` cls for synalm

### LINKED ISSUE #700: passing `None` cls to `synalm`
It does not actually appear possible to pass `None` cls to `synalm` for no correlations:

```py
>>> hp.synalm([[1.], [1.], None, [1.], None, None], new=True)
AttributeError                            Traceback (most recent call last)
...
    444         # Only one spectrum
    445         if lmax is None or lmax < 0:
--> 446             lmax = cls.size - 1
    447         if mmax is None or mmax < 0:
    448             mmax = lmax

AttributeError: 'list' object has no attribute 'size'
```

The reason appears to be that `is_seq_of_seq` does not handle the `None` special case:

```diff
 def is_seq_of_seq(o):
     if not is_seq(o):
         return False
     for s in o:
-        if not is_seq(s):
+        if s is not None and not is_seq(s):
             return False
     return True
```

ntessore:
This PR makes it so that `synalm` accepts Cls that are `None` at the Python layer. Closes #700.

```py
>>> hp.synalm([[1.], None, None, None, None, None], new=True)
array([[-0.14311095+0.j],
       [ 0.        +0.j],
       [ 0.        +0.j]])
```

### DIFF
--- a/healpy/cookbook.py
+++ b/healpy/cookbook.py
@@ -18,7 +18,7 @@ def is_seq(o):
     return hasattr(o, "__len__")
 
 
-def is_seq_of_seq(o):
+def is_seq_of_seq(o, allow_none=False):
     """Check if the object is a sequence of sequences. No check is done on
     the sizes of the sequences.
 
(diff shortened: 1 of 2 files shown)


zonca:
thanks!

PR #720, opened 2021-10-07
docs: improve docs on datapath for pixel weights

zonca:
@ntessore following https://github.com/healpy/healpy/pull/595#issuecomment-935792362

`healpy` has already this functionality, I improved the docs, please review them,
can you also make a test to make sure it works as advertised?

We also have a unit test on this: https://github.com/healpy/healpy/blob/c632a318fecc3ae020f6b5367e024e8927c4df4c/healpy/test/test_pixelweights.py#L37-L52

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,6 @@
 Unreleased
 
+* explain how to create a local datapath for pixel weights <https://github.com/healpy/healpy/pull/720>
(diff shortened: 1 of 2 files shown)


zonca:
@ntessore can you please take a look at this?

ntessore:
Sorry @zonca I didn't see the original notification for this!

I think this is a good enhancement, and shows users that its relatively easy to set up their own datapath. The script works, but the docstring is missing a "cd healpy-data/" between the clone and the script.

All in all :+1:

PR #727, opened 2021-11-04
Update HEALPix from SVN r1192 for HEALPix 3.81

zonca:


lpsinger:
Does the minimum version of healpix-cxx in setup.py need to be updated? What is the minimum version of healpix-cxx that is compatible with healpy?

zonca:
> Does the minimum version of healpix-cxx in setup.py need to be updated? What is the minimum version of healpix-cxx that is compatible with healpy?

yes, I fixed that

lpsinger:
Has this version been released? I'd feel better about updating to a released version.

zonca:
Eric is waiting on me to package the next version of healpy so that he can
release healpix and bundle healpy.

zonca:
The minimum version that I put in setup is 3.80 and that has already been
released.
Eric  will release 3.81

zonca:
see f0158e8

lpsinger:
LGTM.

PR #728, opened 2021-11-04
Update Github actions machinery

zonca:
- remove Travis-CI badge
- [ci] upgrade cibuildwheel action
- [ci] disable delocate workaround


zonca:
@mreineck it seems there is an error configuring HEALPix.
cfitsio is built fine, but then:

```
     checking for cfitsio... yes
    checking for ffgnrwll in -lcfitsio... no
    configure: error: could not find the cfitsio library
```

full log is at https://github.com/healpy/healpy/runs/4466743699?check_suite_focus=true

mreineck:
`ffgnrwll` is a function for accessing "large files" (i.e. > 2GB). It only exists in `libcfitsio.a` if it was configured with the flags `-D_LARGEFILE_SOURCE` and `-D_FILE_OFFSET_BITS=64`. . (see also https://heasarc.gsfc.nasa.gov/fitsio/c/c_user/node32.html). As far as I know, this is the case for all "prebuilt" versions in Linux distributions, but as far as I can tell, these flags are missing in the `configure` invocation in your script.

mreineck:
I would suggest to install the binary `cfitsio` package coming with the distribution in question, instead of building it from scratch and make it a prerequisite of `healpy`. Or does this not work for some reason?

setup.py
```diff
@@ -5,17 +5,17 @@
 import fnmatch
 import sys
 import shlex
-from Cython.Distutils import build_ext
```
lpsinger, line 8: What does this change do?

lpsinger (changes requested):

zonca:
> `ffgnrwll` is a function for accessing "large files" (i.e. > 2GB). It only exists in `libcfitsio.a` if it was configured with the flags `-D_LARGEFILE_SOURCE` and `-D_FILE_OFFSET_BITS=64`. . (see also https://heasarc.gsfc.nasa.gov/fitsio/c/c_user/node32.html). As far as I know, this is the case for all "prebuilt" versions in Linux distributions, but as far as I can tell, these flags are missing in the `configure` invocation in your script.

it seems it is setting LARGEFILE_SOURCE to 1:

gcc -pthread -c -o ./buffers.o -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Dg77Fortran -fPIC  -DPACKAGE_NAME=\"\" -DPACKAGE_TARNAME=\"\" -DPACKAGE_VERSION=\"\" -DPACKAGE_STRING=\"\" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -DSTDC_HEADERS=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_MEMORY_H=1 -DHAVE_STRINGS_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_UNISTD_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_MATH_H=1 -DHAVE_LIMITS_H=1 -D_LARGEFILE_SOURCE=1 -D_FILE_OFFSET_BITS=64 -DHAVE_FTRUNCATE=1 -DHAVE_LONGLONG=1 -DHAVE_SHMEM_SERVICES=1 -DHAVE_FMEMOPEN=1 -DHAVE_NET_SERVICES=1 buffers.c

in:
https://github.com/healpy/healpy/runs/4466743699?check_suite_focus=true#step:3:521

zonca:
> I would suggest to install the binary `cfitsio` package coming with the distribution in question, instead of building it from scratch and make it a prerequisite of `healpy`. Or does this not work for some reason?

we also bundle `cfitsio` with `healpy`, so it would be nice if we also test that building from source works. Still, I'll make a test with that and see if it works.

lpsinger:
> I would suggest to install the binary `cfitsio` package coming with the distribution in question, instead of building it from scratch and make it a prerequisite of `healpy`. Or does this not work for some reason?

I think that Astropy wheels tend to have cfitsio built from bundled source. We're more likely to be compatible with Astropy if we continue to bundle it.

setup.py
zonca: Related to:

https://github.com/healpy/healpy/pull/686

I was getting that error again, so I had to change that. I added reference to the commit now.

zonca:
done in https://github.com/healpy/healpy/pull/740

PR #731, opened 2021-12-29
minor

antolonappan:
check beam_window is array or list.


zonca:
thanks!

zonca:
I need to fix the continuous integration before I can merge this

antolonappan:
> thanks!

No problem, For a recent usage of hp.smoothing, I accidentally gave 'beam_window = True' and the module work without error. So while debugging the wrong final result, I came across this issue.

zonca:
See #744

Issue #724, opened 2021-10-22
Graticule rotation of projview


bhhsieh:
Hi, I am using projview and I want to rotate the map together with the graticule just like what mollview does.
But it seems that projview doesn't rotate the graticule automatically.
Are there any way to do it?

```
hp.mollview(m,rot=(theta,phi,0))
hp.graticule(color='w')

projview(
    m,
    rot=(theta,phi,0),
    graticule=True,
    graticule_labels=True,
    unit="cbar label",
    xlabel="longitude",
    ylabel="latitude",
    cb_orientation="vertical",
    projection_type="mollweide",
    title="Mollweide projection, astro convention (default)",    
)
```
![截圖 2021-10-23 07 05 12](https://user-images.githubusercontent.com/34892184/138528042-707a8b5d-449f-47d5-b542-7a2719f7b5c3.png)

zonca:
@F-Tomas would you like to take a look at this issue?

F-Tomas:
> @F-Tomas would you like to take a look at this issue?

Yes, I will look at it.

F-Tomas:
I believe that this issue can be also closed.

zonca:
Closed in #751

PR #732, opened 2022-01-01
Implementing the rotatable graticule.

F-Tomas:
This commit should solve the issue "Graticule rotation of projview #724". When the map is rotated, a rotated graticule can be turned on by using _rot_graticule=True_.  Also some properties of this graticule can be set, for instance color, linestyle, linewidth etc. The rotated graticule can be used along with the "normal" _graticule graticule=True_. I can add example of this to the JN.


zonca:
rebased on main to fix CI

Issue #676, opened 2021-03-30
Suggestion for experimenting: improved map analysis


mreineck:
The code below implements a new approach at map analysis which should be more intuitive to use and also faster in many cases. Instead of specifying iterations, the user rather specifies a tolerance (measured as the L2 norm of the residual a_lm divided by the L2 norm of the result), and the algorithm iterates until this threshold is reached (or an iteration limit is reached, which can also be specified by the user).
As iteration scheme, the conugate gradient method is employed, which is much more efficient than the one currently used by all Healpix implementations. Its only drawback is that it needs more temporary variables and therefore consumes somewhat more memory during iteration; I have tried to limit this as much as possible, which makes the code look a bit awkward in places.
I'm not using an off-the-shelf algorithm from e.g. `scipy` to demonstrate that the entire algorithm is fairly compact and can easily be implemented in many programming languages if necessary.

Comments very welcome!

```
def hp_map_analysis(map, lmax, mmax, tol=1e-10, maxiter=20):
    alm = healpy.map2alm(map, lmax, mmax, iter=0)
    almnorm = np.vdot(alm,alm).real
    nalm = alm.shape[0]
    nside = healpy.npix2nside(map.shape[0])
    op = lambda x: healpy.map2alm(healpy.alm2map(x, nside, lmax, mmax),
                                  lmax, mmax, iter=0)

    # check if we are fine with just standard map2alm
    # (tol value still needs tuning!)
    if lmax <= 2*nside and tol>=1e-3:
        return alm

    r = op(alm)
    r *= -1
    r += alm
    rdotr = np.vdot(r,r).real
    print(np.sqrt(rdotr/almnorm))
    if rdotr <= tol**2*almnorm:
        return alm

    p = r.copy()
    iter = 1
    while iter<maxiter:
        iter += 1
        apk = op(p)
        alpha = rdotr/np.vdot(p, apk).real
        apk *= alpha
        r -= apk
        del apk
        p *= alpha
        alm += p
        rdotr_old = rdotr
        rdotr = np.vdot(r,r).real
        print(np.sqrt(rdotr/almnorm))
        if rdotr <= tol**2*almnorm:
            return alm
        beta = rdotr/rdotr_old
        p *= beta/alpha
        p += r

    return alm
```

zonca:
Thanks, I think it would be useful to look at a version using `scipy`

mreineck:
Sure, here it is:
```
[...]

PR #734, opened 2022-01-12
Draft: least squares map analysis

mreineck:
This aims to implement the alternative approach to Healpix map analysis discussed in #676 .
Please feel free to push improvements to documentation and suggested code changes!
Most specifically I may need help dealing with support for multiple maps and polarization.

### DIFF
--- a/healpy/sphtfunc.py
+++ b/healpy/sphtfunc.py
@@ -385,9 +385,8 @@ def map_analysis_lsq(maps, lmax, mmax, pol=True, tol=1e-10, maxiter=20):
         return res.reshape((-1,))
 
     def m2a2(x):
-        talm = map2alm(x.reshape((ncomp, -1)), lmax=lmax, iter=0, pol=pol) * (
-            (12 * nside ** 2) / (4 * np.pi)
-        )
+        talm = map2alm(x.reshape((ncomp, -1)), lmax=lmax, iter=0, pol=pol)
+        talm *= (12 * nside ** 2) / (4 * np.pi)
         res = alm2realalm(talm)
         return res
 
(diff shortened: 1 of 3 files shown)


mreineck:
The current version should work for single temperature-only maps. Next planned step is to deal with an IQU-tuple.

mreineck:
Polarization and multiple maps should now work ... I hope I'm doing the correct conversions everywhere.

I think this is now mature enough to discuss naming, interface and documentation details etc. @zonca, any suggestions?

mreineck:
Sorry for the many spurious source changes, those were made by `black`.

mreineck:
@zonca I just remembered that you suggested adding this to `anafast` as an alternative solver, but since the interfaces are rather different (`anafast` does not take a tolerance, and it doesn't return information about convergence) I'm not sure that would be very practical.

mreineck:
Any suggestion how to proceed with this?

zonca:
It seems good to merge for me,
Can you rebase on main and make sure unit tests pass?

zonca (approved):

mreineck:
I have rebased, but it still says "3 commits"; hope that is OK.

zonca:
good that tests pass,
what about calling the function `map2alm_lsq`? "analysis" is not clear what it means to non-experts

mreineck:
I can do this. My concern with `map2alm` is that it is not really clear from the name which operation is meant: it can be the adjoint of `alm2map` or (some approximation to) its inverse. But I'm still using `map2alm` in my own codes for "map analysis" so this will be at least somewhat consistent.

healpy/sphtfunc.py
```diff
@@ -299,6 +299,115 @@ def map2alm(
...
+
+    Returns
+    -------
+    alm : numpy.ndarray(complex)
+        The reconstructed a_lm coefficients
+    rel_res : float
+        The norm of the residual map (i.e. `map-alm2map(alm)`), divided by the
+        norm of `maps`. This is a measure for the fraction of the map signal that
+        could not be modeled by the a_lm
+    iter : int
```
zonca, line 346: `iter` is a reserved word in Python, what about using `n_iter` instead?

PR #735, opened 2022-01-13
Draft: Add a convenience function to compute a_lm of a circular Gaussian

### LINKED ISSUE #733: Add convenience function producing Gaussian beam a_lm?
Maybe I'm not looking in the right places, but I don't think that `healpy` has a function like

`def blm_gauss (fwhm, lmax, pol=False)`

which simply produces a_lm of a circular Gaussian beam. (I know there is one for producing a Gaussian window function, but that's slightly different.)  If this is really missing, would you like me to provide one? I think it could be a useful helper function.

mreineck:
Fixes #733


zonca:
rebased on main to fix CI

healpy/sphtfunc.py
```diff
@@ -1277,3 +1275,46 @@ def check_max_nside(nside):
...
+
+    for l in range(0, lmax + 1):
+        blm[0, Alm.getidx(lmax, l, 0)] = np.sqrt((2 * l + 1) / (4.0 * np.pi)) * np.exp(
+            -0.5 * sigmasq * l * l
+        )
+
+    for l in range(2, lmax + 1):
+        val = np.sqrt((2 * l + 1) / (32 * np.pi)) * np.exp(-0.5 * sigmasq * l * l)
+        blm[1, Alm.getidx(lmax, l, 2)] = val
+        blm[2, Alm.getidx(lmax, l, 2)] = 1j * val
```
zonca, line 1318: you are calling `Alm.getidx(lmax, l, 2)]` twice, can we do just Q inside the loop and then after the loop something like:

     blm[2] = 1j * blm[1]

zonca (changes requested):
It would be nice to have a unit test with a very simple case.

healpy/sphtfunc.py
mreineck: You're absolutely right! Sorry, that comes from relying too much on optimizing compilers. Will fix it soon.

Anyway, I'm aware that this particular implementation is painfully slow. It can be made much faster, but will also become somewhat obscure in the process. Given that there is not much computation in this routine  (only O(lmax)), I thought we can get away with the slow version that looks nicer :-)

mreineck: Regarding a unit test,I'm not sure yet how to design one, but I'll think about it some more.

zonca:
@mreineck maybe you can run a simple example like this in fortran or C++ and we compare the arrays?

https://gist.github.com/35622edf0ccb6622907682280ad2296b

mreineck:
I'll hopefully have time to do that tomorrow!

mreineck:
If a pure regression test is OK, then I think I'm done.

zonca:
moved `black` formatting to main

zonca (approved):

PR #739, opened 2022-01-14
Use cibuildwheel's default manylinux image

lpsinger:
We had pinned it to manylinux1, which is past end of life now.


zonca:
yes, I was trying something similar in https://github.com/healpy/healpy/pull/728 with no success

PR #740, opened 2022-01-14
Several changes to fix building of wheels

lpsinger:


zonca:
Thank you very much for fixing this

zonca:
@lpsinger I just noticed that the linux one didn't pass, it actually timed out after 6 hours:

https://github.com/healpy/healpy/runs/4813376952?check_suite_focus=true

lpsinger:
Yeah, it's building lots of wheels serially. I'll research what is the best way to parallelize cibuildwheel in GitHub Actions.

zonca:
it seems the problem is the testing on one of the wheels, building and testing the wheels takes 5 minutes each

PR #742, opened 2022-01-14
Only build wheels for 64-bit archs (skip i686)

lpsinger:
The cibuildwheel CI job is timing out. Skip wheels for 32-bit archs (i686). These archs are very rare nowadays and it is not common for our upstream dependencies to provide wheels for them anyway. Numpy, for example, does not distribute i686 wheels.

This should significantly reduce the build time because the CI job will only have to build half as many wheels. The skipped i686 wheels were the slowest anyway, because it had to build wheels for many of the dependencies too (including Numpy).

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -41,6 +41,7 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.3.1
         env:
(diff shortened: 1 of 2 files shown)


lpsinger:
Failing due to pypa/cibuildwheel#992.

lpsinger:
The wheels are all building now. @zonca, please review

Issue #747, opened 2022-01-28
norm='log' or 'hist' does not work in hp.projview: NotImplementedError


skoldobskiy:
Hello!

It looks like the normalization of the colorbar is currently not implemented.
I don't know if I should report that or if you already know about it.

Thanks

zonca:
https://github.com/F-Tomas would you like to take a look into this?

F-Tomas:
Yes, will check it.

zonca:
@F-Tomas `log` in particular would be useful to have

zonca:
implemented in #751

PR #751, opened 2022-03-21
Adding the norm feature and a cbar_ticks argument.

F-Tomas:
Solving issue #747


zonca:
thanks @F-Tomas !

trygvels:
It would be great if this also included symlog functionality! With a linear threshold between some specified value, perhaps defaulted to 1.

F-Tomas:
I will put that on my to-do list.

Issue #756, opened 2022-03-31
Margins applied twice with subplots for `cartview` and `gnomview`


mpetroff:
While working on #755, I noticed that the margins appear to be applied twice with subplots for `cartview` and `gnomview`.

When `sub != None` for `cartview` and this `else` statement is run
https://github.com/healpy/healpy/blob/215c747b4bf94631dce0f44a9f04d1d778581382/healpy/visufunc.py#L773-L795
the margins are applied to `extent` in the `else` statement but then immediately applied a second time right after. This is also the case for `gnomview`.

Unless I'm misinterpreting this code, it appears to be a bug. If so, it raises the question as to what should be done, since it changes existing behavior and could potentially break existing plotting scripts.

zonca:
thanks @mpetroff, it seems like a bug to me as well, have you tested the code just to make sure that it really doubles the margin?

About fixing this, it seems the easiest would be to remove the application of margins inside the `else`, right? so this applies only to scripts using `subplots`

mpetroff:
I hadn't tested it before, but now I did. The margins are definitely applied twice. With `hp.cartview(np.random.rand(hp.nside2npix(8)), sub=(1, 1, 1), margins=(0.24, 0.24, 0.24, 0.24))`, there's a extremely tiny map in the center of the plot. If `margins` is changed to `(0.25, 0.25, 0.25, 0.25)`, there's no map, and there are runtime warnings, probably due to division by zero.

The issues with margins are actually worse than I originally thought. The margin behavior across all five projections is the same for subplot syntax, with the margins doubled. For `mollview`, `orthview`, and `azeqview`, the `margins` parameter doesn't do anything when subplot syntax isn't used.

mpetroff:
Fixed in #757.

PR #757, opened 2022-04-01
Fix plot margins

mpetroff:
See further discussion in #756.

For `mollview`, `orthview`, and `azeqview`, the `margins` parameter previously did not do anything when subplot syntax was not used.

For `mollview`, `gnomview`, `cartview`, `orthview`, and `azeqview`, the `margins` parameter was previously double applied when subplot syntax was used.

This PR fixes both issues. Note that with the change in margin behavior, the output of existing scripts that use the `margins` parameter may change.


zonca (approved):
Thanks!

can you please add a line about this in [CHANGELOG.rst](https://github.com/healpy/healpy/blob/main/CHANGELOG.rst)?

Issue #760, opened 2022-04-26
Projview AttributeError: 'MollweideAxesSubplot' object has no attribute 'proj'


mfouesneau:
How do you annotate those projections? the `proj` attribute is part of the older views but is not defined in `projview`

PR #752, opened 2022-03-23
New features added to Projview

trygvels:
Hi!
I have added some missing features to the projview function. 

The new additions are
- subplots
- remove mono-dipole
- top left and right labels
- preserving tickmarkers when specifying ticklabels
- fontstyle
- tickdirection and vertical label rotation
- format argument now used in labeling

I have also made som minor changes to existing functionality, but its up to you if you want to keep it
- Adjusted rotated graticule to match standard graticule
- Added a width argument, because overriding_plot_properties was a bit of a hassle

An example of the new features is:
```Python
hp.projview(dust, rlabel=r"A$_{\mathsf{dust}}$", llabel=r"$Q$", unit=r"$\mu$K", fontname="serif", min=-100, max=100, width=10, show_tickmarkers=True, cbar_ticks=[-100,-50, 0,50, 100], cb_orientation="vertical", sub=221, override_plot_properties={"cbar_tick_direction":"in"})
hp.projview(dust, norm="symlog", rlabel=r"A$_{\mathsf{dust}}$", llabel=r"$Q$", unit=r"$\mu$K", cbar_ticks=[-100,-50, 0,50, 100], remove_mono=True, show_tickmarkers=True,  cb_orientation="horizontal", sub=222, override_plot_properties={"cbar_tick_direction":"in"})
hp.projview(cmb,  rlabel=r"A$_{\mathrm{cmb}}$", llabel=r"$I$",  unit=r"$\mu$K",  min=-3e3, max=3e3,  cb_orientation="vertical", sub=223, graticule=True, )
hp.projview(cmb,  rlabel=r"A$_{\mathrm{cmb}}$", llabel=r"$I$", unit=r"$\mu$K", fontname="serif",  remove_dip=True, min=-300, max=300,  cb_orientation="horizontal", sub=224, graticule=True, rot_graticule=True, rot=(0,0,0))
```
Which produces:
example

**_Missing features:_**
- Currently, there is no way to use the same symlog functionality as mollview. I suspect this is a simple fix.
- I started implementing a coordinate rotation of the custom graticule, but ran out of time. This would be useful.

PR #761, opened 2022-05-02
Fixed Unused parameters and other bugs

### LINKED ISSUE #713: Unused parameter "cbar" - Can't hide colorbar in new projview
Hi!
Looks like the 'cbar' boolean parameter never gets used in the projview function, so there is no way to hide the colorbar currently.

mfouesneau:
This should fix #713  & #760 

The main changes are:
* cleaning unused variables, imports
* more consistent argument/keywords names (also with with matplotlib)
* `newproj` cbar keyword now considered
* `newproj` norm now considered and expended to other norms (including `histeq`, and `matplotlib.Normalize` objects)
* `newprojplot` now includes `lonlat` and `ax`  keywords
* `newprojtext` is a new function that allows text (a wrapper around plt.text) which also allows `lonlat`

### DIFF
--- a/healpy/newvisufunc.py
+++ b/healpy/newvisufunc.py
@@ -1,13 +1,15 @@
 __all__ = ["projview", "newprojplot"]
 
+import warnings
 import numpy as np
-from .pixelfunc import ang2pix, npix2nside
-from .rotator import Rotator
-from .projaxes import get_color_table
 import matplotlib.pyplot as plt
 from matplotlib.projections.geo import GeoAxes
-from matplotlib.ticker import MultipleLocator, FormatStrFormatter, AutoMinorLocator
-import warnings
+from matplotlib.ticker import MultipleLocator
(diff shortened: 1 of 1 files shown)


zonca:
great contribution thanks!
can you please update accordingly the example notebook https://github.com/healpy/healpy/blob/main/doc/newvisufunc_example.ipynb?
which is probably causing the readthedocs build to fail.

zonca:
apologies, I first need to merge #752, which is also related to `projview`

zonca:
ok, I merged #752, if you have time to rebase your changes on the current `main` it would be really helpful, otherwise I'll try to do myself in the future.

mfouesneau:
#752 has conceptual divergence with the original function, I think it would be nice to let the user deal with subplots, color bar, etc, and only provide the equivalence to the previous methods.

Can I make executive decisions?

PR #777, opened 2022-08-20
Use RING for max_pixrad as NEST only allows power of two nsides

chpatrick:


zonca:
@mreineck what do you think about this change?

mreineck:
Looks perfectly reasonable to me. Asking for the maximum pixel radius at arbitrary nside will only work when using RING.

PR #797, opened 2022-10-27
Move projview's subplots_adjust inside check on return_only_data

joezuntz:
Right now the projview function still does one figure-related operation even if return_only_data is set to True. This change stops that from happening by moving that call inside a check.

### DIFF
--- a/healpy/newvisufunc.py
+++ b/healpy/newvisufunc.py
@@ -506,9 +506,6 @@ def projview(
     if graticule and graticule_labels:
         left += 0.02
 
-    if not return_only_data:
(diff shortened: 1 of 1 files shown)


joezuntz:
I think it's probably a mistake that actions building release and publishing to pypi are running on PRs?  You can definitely switch that off.

zonca:
> I think it's probably a mistake that actions building release and publishing to pypi are running on PRs? You can definitely switch that off.

No, it is for testing builds on different plaforms, PyPI is skipped for non-releases

PR #819, opened 2023-01-11
Build wheels for aarch64 under emulation

### LINKED ISSUE #818: 'pip install' fails on ARM platforms, including Apple Silicon
So here's a Dockerfile:
```Dockerfile
FROM ubuntu:jammy

RUN apt-get update
RUN apt-get install python3 pip -y
RUN pip install healpy
```
(note that this will install python3.10, today)

Try `docker build` with that, and you'll get an error:

```
#7 39.94         File "/usr/lib/python3.10/subprocess.py", line 369, in check_call
#7 39.94           raise CalledProcessError(retcode, cmd)
#7 39.94       subprocess.CalledProcessError: Command '['/bin/sh', '/tmp/pip-install-m3p7snpr/healpy_d012e8a7590242d9b4afdb9a97bef27c/healpixsubmodule/src/cxx/configure', '--prefix=/tmp/pip-install-m3p7snpr/healpy_d012e8a7590242d9b4afdb9a97bef27c/build/temp.linux-aarch64-3.10', '--disable-shared', '--with-pic', '--disable-maintainer-mode']' returned non-zero exit status 1.
#7 39.94       [end of output]
#7 39.94
#7 39.94   note: This error originates from a subprocess, and is likely not a problem with pip.
#7 39.94   ERROR: Failed building wheel for healpy
#7 39.94 ERROR: Could not build wheels for healpy, which is required to install pyproject.toml-based projects
#7 39.94 Failed to build healpy
```
There's a lot more in the scrollback, but crawling up through it, this seems to be related to the cfitsio redistribution:

```
#7 39.94       checking for cfitsio... yes
#7 39.94       checking for ffgnrwll in -lcfitsio... no
#7 39.94       configure: error: could not find the cfitsio library
```
The first and third lines of that seem to be in some disagreement.

The full output is very long. I've placed it here: https://gist.github.com/spenczar/f49c4a427dcb1bc961d93125e1847efa

lpsinger:
Fixes #818.

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -38,9 +38,17 @@ jobs:
         with:
           submodules: true
 
+      - name: Set up QEMU
+        if: runner.os == 'Linux'
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: arm64
+
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.11.4
         env:
+          # Build wheels for aarch64 under emulation.
+          CIBW_ARCHS_LINUX: "auto aarch64"
           # Skip PyPy builds; Astropy wheels are not built for PyPy.
           # Skip i686 builds.
           # Skip MUSL libc builds.


.github/workflows/cibuildwheel.yml
```diff
@@ -41,6 +41,8 @@ jobs:
       - name: Build wheels
```
spenczar, line 41: I think you'll need a step just before the "Build wheels" step which sets up QEMU for emulation:

```yaml
      - name: Set up QEMU
        if: runner.os == 'Linux'
        uses: docker/setup-qemu-action@v2
        with:
          platforms: arm64
```

lpsinger: Right you are! Thanks.

lpsinger:
> This could be resolved by dropping Python 3.7 support (just for aarch64, I suppose), or by pinning markupsafe to the old version (`markupsafe==2.0.1`).

See #821.

spenczar:
I can confirm that the `aarch64` wheels generated in the most recent run (https://github.com/healpy/healpy/actions/runs/3896259703) work on my M1 Macbook Pro. Thanks!

PR #829, opened 2023-02-19
Fixed colorbar label for newvisfunc (cartesian and geographic proj)

skoldobskiy:
Hi, I propose small changes for `newvisufunc`, which allow one to better locate the colorbar label for `geographic` (for `cb _orientation='vertical'`) and `cartesian` (for both `cb_orientation` cases) projections.
For me it is the first experience with pull request, please let me know if something should be done differently.


zonca:
thanks for your contribution!

PR #856, opened 2023-06-30
Bump macOS version used for wheel builds

lpsinger:
GitHub has retired macOS 10.15 support.


lpsinger:
@zonca, would you please review and merge this? It fixes the macOS build.

PR #862, opened 2023-07-31
Fix absolute imports in .pyx files

lpsinger:
See https://github.com/conda-forge/healpy-feedstock/pull/57.

### DIFF
--- a/healpy/src/_query_disc.pyx
+++ b/healpy/src/_query_disc.pyx
@@ -7,7 +7,7 @@ from libcpp.vector cimport vector
 cimport cython
 
 from _common cimport int64, pointing, rangeset, vec3, Healpix_Ordering_Scheme, RING, NEST, SET_NSIDE, T_Healpix_Base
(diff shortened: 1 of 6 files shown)


zonca (approved):

lpsinger:
@zonca, would you please do a release so that we can fix this in the Conda packages without resorting to a patch?

healpy/src/_query_disc.pyx
```diff
@@ -7,7 +7,7 @@ from libcpp.vector cimport vector
 cimport cython
 
 from _common cimport int64, pointing, rangeset, vec3, Healpix_Ordering_Scheme, RING, NEST, SET_NSIDE, T_Healpix_Base
-from _pixelfunc import isnsideok
+from ._pixelfunc import isnsideok
```
pkgw, line 10: I'm not sure if it actually matters, but note that this would be more properly `.pixelfunc`, not `._pixelfunc`.

PR #865, opened 2023-08-01
Move package metadata to pyproject.toml

lpsinger:
This has several advantages:
- The project metadata is now stored as configuration, not code.
- We don't need to explicitly read the README file.
- The healpy/version.py file is now generated automatically.


lpsinger:
@zonca, my last push should have fixed the build failure due to the syntax error in authors.

zonca (approved):

PR #877, opened 2023-08-17
Build against numpy 1.25, not oldest-supported-numpy

lpsinger:
From Numpy 1.25 onward, Numpy has API backwards compatibility. It is no longer necessary to build against oldest-supported-numpy.


zonca (approved):

zonca (approved):

lpsinger:
Oof, this broke the build. That's OK, my next few PRs will fix it.

PR #881, opened 2023-08-17
Fix Numpy API deprecations

lpsinger:


lpsinger:
@zonca, it looks like this fixes the cibuildwheel jobs. This is ready to review and merge.

PR #885, opened 2023-09-04
Try to make Gaussian blm consistent

### LINKED ISSUE #861: sphtfunc.blm_gauss() is not consistent with the rest of healpy
This is a heads-up that the approximations used to compute the Gaussian beam `a_lm` created by `blm_gauss()` are not consistent with the window functions produced by `gauss_beam()`. I'll try to address this in a few weeks' time, but wanted to draw attention to this already now.
The code in `gauss_beam` follows the Challinor et al. paper, https://arxiv.org/pdf/astro-ph/0008228.pdf, while `blm_gauss` is based on the Fortran "Beam" package from Imperial College (?) that was used inside Planck LevelS.

I would greatly appreciate any comments on this! @hivon, which of the alternatives do you suggest?

mreineck:
Fixes #861

I'd love to get some independent opinions on this, especially on the normalization factor for the polarised components.

@hivon, do you think this is the right expression?

### DIFF
--- a/healpy/sphtfunc.py
+++ b/healpy/sphtfunc.py
@@ -1421,14 +1421,14 @@ def blm_gauss(fwhm, lmax, pol=False):
 
     for l in range(0, lmax + 1):
         blm[0, Alm.getidx(lmax, l, 0)] = np.sqrt((2 * l + 1) / (4.0 * np.pi)) * np.exp(
-            -0.5 * sigmasq * l * l
+            -0.5 * sigmasq * l * (l + 1)
         )
 
     if pol:
         for l in range(2, lmax + 1):
             blm[1, Alm.getidx(lmax, l, 2)] = np.sqrt(
-                (2 * l + 1) / (32 * np.pi)
-            ) * np.exp(-0.5 * sigmasq * l * l)
+                (2 * l + 1) / (4 * np.pi)
+            ) * np.exp(-0.5 * sigmasq * (l * (l + 1) - 4))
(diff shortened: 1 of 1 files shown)


mreineck:
Regression test currently fails, because I'm comparing against hardcoded arrays. I'll adjust those once we have converged to the right formula.

zonca:
@mreineck what about writing a unit test that uses both this and `blm_gauss` to show that they are consistent?
@hivon any feedback on this?

hivon:
Unit tests are always  a good idea.
@zonca, @mreineck I enclose below my reply from Sep 4 about the numerical factors

-------- Forwarded Message --------
Subject:     Re: [healpy/healpy] Try to make Gaussian blm consistent (PR #885)
Date:     Mon, 4 Sep 2023 14:29:24 +0200
From:     Eric Hivon [<hivon@iap.fr>](mailto:hivon@iap.fr)
To:     [reply+ABKVQDMCLQKEQUOTOD7QXQODALZ6HEVBNHHHAD7NIU@reply.github.com](mailto:reply+ABKVQDMCLQKEQUOTOD7QXQODALZ6HEVBNHHHAD7NIU@reply.github.com)



Hi @mreineck,
this is the factor I use for E and B beams, based on the Challinor et al paper.
While it matters for theoretical studies, it won't matter as much for data analysis, where
the temperature and polarization gains and beams are calibrated separately.

Cheers,
Eric

PR #898, opened 2024-02-02
Move cibuildwheel conf to pyproject.toml, fix OpenMP for macOS

lpsinger:
https://github.com/healpy/healpy/pull/896 accidentally switched
the macOS builds form gcc to Xcode's clang; this disabled OpenMP on
macOS because the Xcode's clang does not support the `-fopenmp` flag.

Since gcc was causing crashes on macOS 14 (see
https://github.com/matplotlib/matplotlib/issues/27724), switch from
gcc to homebrew's clang, which does support OpenMP and does not
cause crashes.

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -45,17 +45,13 @@ jobs:
         with:
           platforms: arm64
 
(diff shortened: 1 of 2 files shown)


lpsinger:
@zonca, would you please review?

PR #906, opened 2024-02-23
Shrink wheels by not linking against libcurl

lpsinger:
Cfitsio links against libcurl to support remote file I/O. On Linux, curl adds 24 shared libraries totalling 11 MB in size that get bundled into the wheel.

Nowadays, remote I/O support for FITS reading in Python is not provided by cfitsio but by the Python code in astropy.io.fits.

It's unlikely that the cfitsio remote file I/O was working anyway because the default certificate paths baked into the libraries are unlikely to match their locations on the target system, and there is no way to control those paths from the Python-level API.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,4 +3,4 @@
 	url = https://github.com/healpy/healpixmirror.git
 [submodule "cfitsio"]
 	path = cfitsio
-	url = https://github.com/healpy/cfitsio.git
+	url = https://github.com/HEASARC/cfitsio.git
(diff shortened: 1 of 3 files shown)


lpsinger:
Looks like this needs a little more work.

lpsinger:
@zonca, I think that the build failure on this PR will be fixed by updating the bundled cfitsio sources to the latest stable version. Would you please update cfitsio?

Background: in the older version of cfitsio that we have, the flag `-lcurl` is hardcoded in the pkgconfig file. It is no longer hardcoded in the most recent version; it is added only when cfitsio is built with curl support.

lpsinger:
Note: merge #908 first.

zonca (approved):

PR #908, opened 2024-02-27
Update bundled cfitsio to 4.3.1, use official GH mirror

lpsinger:
- Update bundled cfitsio to 4.3.1.
- Use official GitHub mirror maintained by HEASARC instead of our own mirror.


zonca:
do you think this should be merged even with the error in `macos-14`?

lpsinger:
The error actually appears to be fixed by omitting libcurl, so you could just merge #906 instead.

Issue #907, opened 2024-02-23
Make bulky dependencies optional


swyatt7:
Hello,

We (me, @lpsinger, and @ jak574) were discussing the potential for `healpy` make their bulkier dependencies (e.g. `scipy` and `matplotlib`) optional for installs. Our application has ran into an issue where including `healpy`  has increased the dependency size by  ~250MB, and are running into limits with our cloud deployment. 

We were considering maybe exploring the options of having something like:

`pip install healpy['all']` includes all of the current dependencies,

and 

`pip install healpy` will install the package sans `scipy` and `matplotlib`,

or conversely

`pip install healpy['lite']` will install the package sans `scipy` and `matplotlib`. 

Obviously a change like this will have to take into consideration existing applications that rely on `healpy` retaining those dependencies, and this proposal will potentially break them. 

Let me know if the `healpy` team will consider making a change like this, and I'll put in a PR for it. All that it will require will be to move the initial file level imports into the function-level, which will be easy for `scipy` where it is only dependent in the `sphtfunc.py`, and a little more tedious for `matplotlib`.

Best,

-SW

lpsinger:
> `pip install healpy['lite']` will install the package sans `scipy` and `matplotlib`.

Extras cannot subtract dependencies; they can only add them.

zonca:
I think it is fine if we have `pip install healpy` without `scipy` and `matplotlib` and `pip install healpy['all']` includes all of the current dependencies.

lpsinger:
Fixed by #910.

PR #910, opened 2024-03-01
Optional dependencies

swyatt7:
# Changes
Made `scipy` and `matplotlib` optional dependencies, where they can be included by doing `pip install "healpy[all]"`.  Addresses https://github.com/healpy/healpy/issues/907

# Implications
The `scipy` dependency was simple, and moving it to the function level was trivial. However for `matplotlib`, custom `healpy` classes inherit directly from `matplotlib` classes, thus simply moving the imports to the function level wasn't possible. In order to have a _semi-stable_  library, I had to remove the `visfunc`, `zoomtool`, and `newvisfunc` imports from `healpy/__init__.py`. Which means that users will have to call those methods by descending into those modules directly (e.g. `healpy.visfunc.mollview(...)`).

- Upsides: 
  - Importing `healpy` will be much faster.
  - Installing sans `scipy` and `matplotlib` will be much slimmer.
- Downsides: 
  - Will break existing applications where plotting methods are used.
  - API documentation and examples will certainly be affected.

### DIFF
--- a/healpy/__init__.py
+++ b/healpy/__init__.py
@@ -94,10 +94,29 @@ from ._pixelfunc import ringinfo, pix2ring
 from ._sphtools import rotate_alm
 from .rotator import Rotator, vec2dir, dir2vec
 from ._healpy_pixel_lib import UNSEEN
+try:
+    from .visufunc import (
+        mollview,
+        graticule,
+        delgraticules,
+        gnomview,
+        projplot,
+        projscatter,
+        projtext,
+        cartview,
+        orthview,
+        azeqview,
+    )
+    from .zoomtool import mollzoom, set_g_clim
+    from .newvisufunc import projview, newprojplot
+except:
+    pass
+
 from .fitsfunc import write_map, read_map, read_alm, write_alm, write_cl, read_cl
 from ._masktools import dist2holes_healpy as dist2holes
 from ._hotspots import hotspots_healpy as hotspots
 from ._line_integral_convolution import line_integral_convolution
+
 
 from .utils.deprecation import deprecated
 
--- a/healpy/__init__.py
+++ b/healpy/__init__.py
@@ -94,25 +94,25 @@ from ._pixelfunc import ringinfo, pix2ring
 from ._sphtools import rotate_alm
 from .rotator import Rotator, vec2dir, dir2vec
 from ._healpy_pixel_lib import UNSEEN
-from .visufunc import (
-    mollview,
-    graticule,
-    delgraticules,
-    gnomview,
-    projplot,
-    projscatter,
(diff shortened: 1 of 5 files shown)


lpsinger (changes requested):
Please fix the Readthedocs build failure.

@jak574 suggests:

> Did you consider having conditional imports in `__init__.py` based on whether matplotlib is available? Simpliest way would be to have a try/except with import matplotlib at the top, followed by all the imports that are dependent on matplotlib. I'm sure there are better ways also.

healpy/__init__.py
```diff
@@ -94,26 +94,29 @@
...
+        projscatter,
+        projtext,
+        cartview,
+        orthview,
+        azeqview,
+    )
+    from .zoomtool import mollzoom, set_g_clim
+    from .newvisufunc import projview, newprojplot
+except:
+    pass
```
lpsinger, line 113: If the import fails for any reason other than Matplotlib being missing, then this should raise an exception.

```suggestion
try:
    import matplotlib
except ImportError:
    pass
else:
    from .visufunc import (
        mollview,
        graticule,
        delgraticules,
        gnomview,
        projplot,
        projscatter,
        projtext,
        cartview,
        orthview,
        azeqview,
    )
    from .zoomtool import mollzoom, set_g_clim
    from .newvisufunc import projview, newprojplot
```

lpsinger (changes requested):

healpy/visufunc.py
```diff
@@ -63,8 +63,6 @@
 
 from . import projaxes as PA
 import numpy as np
-import matplotlib
-import matplotlib.pyplot as plt
```
lpsinger, line 67: Due to cd9818aeb74e879c267ca8dcae66eca449a8d1a4, you no longer need to move these imports from module to function scope.

lpsinger (changes requested):

healpy/visufunc.py
swyatt7: fixed in recent push

lpsinger (approved):

lpsinger:
Looks good to me. @zonca?

zonca (approved):
it looks good, can you please also add a line in `CHANGELOG.rst` under "Unreleased"?

zonca:
also, how would this work for our `conda` package?

lpsinger:
> also, how would this work for our `conda` package?

Conda doesn't have optional dependencies (see https://github.com/conda/conda/issues/2984). The Conda package would still depend on both scipy and matplotlib.

lpsinger:
@zonca, would you please do a release to include this PR, #908, #905, and #896?

PR #917, opened 2024-03-07
update C++ sources to SVN commit 1238

zonca:


zonca:
@mreineck @lpsinger I updated from SVN but it looks like there is a compilation problem.
Here are the differences between the 2 sources:
https://github.com/healpy/healpixmirror/pull/3

mreineck:
I have no idea what's happening here, but looking at the diffs, it must be related to the deletion of the `libz` block in `configure.ac`. This change was done as an attempt to fix https://sourceforge.net/p/healpix/bugs/137/ ...

@lpsinger, do you have any advice?

lpsinger:
> I have no idea what's happening here, but looking at the diffs, it must be related to the deletion of the `libz` block in `configure.ac`. This change was done as an attempt to fix https://sourceforge.net/p/healpix/bugs/137/ ...

That _was_ committed to trunk: https://sourceforge.net/p/healpix/code/1229/

Was that unintentional?

mreineck:
No, it was intentional, since it addressed the reported problem and I could not really understand why we should look for a `zlib` that we don't directly depend on. Imagine a situation (e.g. on Windows), where `zlib` is statically linked into `libcfitsio`. Configuration should still work in that case.

Is there a way to write `configure.ac` which wll work in all cases?

lpsinger:
> Is there a way to write `configure.ac` which wll work in all cases?

This might work. I don't know. What did you want me to look at when you said:

> I have no idea what's happening here, but looking at the diffs, it must be related to the deletion of the libz block in configure.ac.

lpsinger:
Oh, I see. You're asking why the build is failing? This is a bug in cfitsio; it ought to advertise the flags for linking against libz in its pkgconfig file.

mreineck:
So we are between a rock and a hard place ... `configure` will potentially fail, no matter whether I put the `zlib` block in or leave it out ... I have no idea what I should do :-/

@lpsinger, I see that you are active on `libcfitsio`'s issue tracker - do you expect that `cfitsio` will fix this in the foreseeable future?

lpsinger:
> @lpsinger, I see that you are active on `libcfitsio`'s issue tracker - do you expect that `cfitsio` will fix this in the foreseeable future?

I'm not sure. The fix is straightforward, but HEASARC seems to be pretty new to GitHub and still getting the hang of accepting outside contributions.

zonca:
I don't see an open issue or PR on https://github.com/HEASARC/cfitsio for this. Can you open one?
They do not seem responsive on Github, what if we fork `cfitsio` under `healpy` and we apply the fix there?

esabol:
> I don't see an open issue or PR on https://github.com/HEASARC/cfitsio for this. Can you open one? They do not seem responsive on Github [...]

That's not true. HEASARC staff has responded to nearly every open issue at https://github.com/HEASARC/cfitsio/issues within a few days of them being opened.

> [...] what if we fork `cfitsio` under `healpy` and we apply the fix there?

Please don't fork.

lpsinger:
> > I don't see an open issue or PR on https://github.com/HEASARC/cfitsio for this. Can you open one? They do not seem responsive on Github [...]
> 
> That's not true. HEASARC staff has responded to nearly every open issue at https://github.com/HEASARC/cfitsio/issues within a few days of them being opened.
> 
> > [...] what if we fork `cfitsio` under `healpy` and we apply the fix there?
> 
> Please don't fork.

Of course not, let's work with the HEASARC team to fix this. They work right down the hall from me, after all!

> I don't see an open issue or PR on https://github.com/HEASARC/cfitsio for this. Can you open one?

I took leave today to spend some time with a friend who was visiting from out of town, so I have been away from my computer. I will open an issue or PR for cfitsio by early this coming week. Briefly, my proposed solution will be to rework the Makefile to use automake and libtool rather than to do the linking "by hand" as it does now. That should result in generating correct linking arguments for dependent packages regardless of whether the build is static or shared.

trjaffe:
To clarify, indeed in the past, this repo was not well monitored, and some PRs have been left to languish.  We are working to improve this situation and hope very much that with your help, this will become an active repo with your contributions accepted, if not speedily, then in a reasonable delay.  Thanks Leo!  Hi Martin!

mreineck:
Hi Tess! :-)

@lpsinger, I wonder: would it be good to have an option (say, `ADD_ZLIB_FLAGS`) in `Healpix_cxx`'s `autoconf` machinery to work around the problem until a new `libcfitsio` version comes out? Basically the

```
PKG_CHECK_MODULES([ZLIB], [zlib], , [ZLIB_LIBS=-lz])
CFLAGS="$CFLAGS $ZLIB_CFLAGS"
LIBS="$LIBS $ZLIB_LIBS"
```

we originally had in there, just put behind an optional `configure` flag?

I'd be fine with this kind of change, but my `autoconf` is a little rusty, so I'm not sure I'll get it right at first try.

lpsinger:
As zlib is not a direct dependency of healpix-cxx, we shouldn't need to detect its presence ourselves at all.

zonca:
@lpsinger imported alice3 patch into healpix here

lpsinger:
The build is still failing, evidently due to the zlib detection issue. @trjaffe, one of your colleagues had mentioned that rewriting that cfitsio Makefile in automake was on their to-do list. Is it still? Or do you need outside help from me to do that?

bryanirby:
Hi - I'm trying to wrap up some priority fixes for HEASoft, but hope to start looking at converting CFITSIO to the automake system tomorrow.  I'm not currently able to provide an ETA, but will prioritize this as much as possible.

To try to address what looks to be some confusion about the CFITSIO GitHub above: a different team put a copy of CFITSIO up several years ago, but it was not managed or maintained by the CFITSIO developer(s).  When we put the current copy up a few months ago, the old, deprecated one was moved aside and renamed as "old_cfitsio".

We have tried to respond in a timely fashion to all PRs & issues in the new repository, but are a small team with other projects competing for our attention.  So, we do very much appreciate everyone's feedback and contributions to CFITSIO.

lpsinger:
@mreineck, I find that this patch applied to the healpix repo fixes the issue. Would you please review and merge it? [0001-Detect-libz-libm-dependencies-of-cfitsio.patch](https://github.com/healpy/healpy/files/14985959/0001-Detect-libz-libm-dependencies-of-cfitsio.patch)

mreineck:
Thanks, @lpsinger!
Not that I mind much, but out of curiosity: does this mean that `libcfitsio` will be linked statically now?
If I read correctly, `PKG_CHECK_MODULES_STATIC` requires `pkgconfig >=0.29`, do we need to test for  this anywhere?

mreineck:
I just realized that pkgconfig 0.29 is *old* ... patch is now applied.

lpsinger:
> Thanks, @lpsinger! Not that I mind much, but out of curiosity: does this mean that `libcfitsio` will be linked statically now? If I read correctly, `PKG_CHECK_MODULES_STATIC` requires `pkgconfig >=0.29`, do we need to test for this anywhere?

What makes cfitsio different from most other autotools libraries is that since it doesn't use automake, it does not have the same behavior about when it builds static vs. shared libraries. So I think that right now this is failing in our CI pipeline because cfitsio _is_ being built as a static library.

mreineck:
Thanks, I'll look into that!

Possible thought for the future: the only functions where `healpy` requires FITS functionality from `healpix_cxx` are `read_weight_ring` and `read_fullweights`. Perhaps we could reimplement these two (small and not performance-critical) functions in `cython` and reduce the external dependencies of `healpix_cxx` drastically.
This would have to happen anyway when switching to `ducc`

PR #931, opened 2024-04-03
Fix config error in readthedocs

zonca:


zonca:
`Config validation error in build.os. Value os not found.`

PR #933, opened 2024-04-10
Build against Numpy>=2.0.0rc1

lpsinger:
Numpy 2.0.0 is coming out soon. In order to support both Numpy 1.x and 2.x, we must build against Numpy 2.x.

The [official guidance from Numpy](https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-abi-handling) is that we should be targeting numpy 2.0.0rc1 as a build dependency now.


lpsinger:
@zonca, would you please review this?

zonca (approved):
very good, it is just a depencency at build time

PR #942, opened 2024-04-23
Ensure that we build cfitsio as a shared library

lpsinger:
cfitsio's `make install` target weirdly does not depend on `make all`, so `make install` by itself doesn't install shared libraries.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -308,7 +308,6 @@ setup(
             {
                 "pkg_config_name": "cfitsio",
                 "local_source": "cextern/cfitsio",
-                "supports_non_srcdir_builds": False,
             },
         ),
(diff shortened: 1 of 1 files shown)


mreineck:
Thanks, should be in SVN now!

lpsinger:
Now the build is failing on macOS because the shared linking in cfitsio's build system is broken. MacPorts has a patch for it: https://github.com/macports/macports-ports/blob/master/science/cfitsio/files/patch-configure.diff

This would be fixed by https://github.com/HEASARC/cfitsio/issues/18.

lpsinger:
@zonca, this is ready for review now.

PR #943, opened 2024-04-23
update C++ sources to r1240

zonca:
@lpsinger


zonca:
do this update as part of #942

PR #953, opened 2024-06-25
Rename trapz function from scipy.integrate module to trapezoid

### LINKED ISSUE #952: scipy trapz function is not supported anymore
Since the very recent release of `scipy 1.14`, the `trapz` function used [here](https://github.com/healpy/healpy/blob/main/lib/healpy/sphtfunc.py#L1339) is now removed (there was some deprecation warning since `scipy` 1.12 but now it's expired see [release notes](https://docs.scipy.org/doc/scipy/release/1.14.0-notes.html#expired-deprecations)).

I see that `healpy` is in the process of releasing a new [version](https://github.com/healpy/healpy/issues/914) but can you add the following basic fix
```python
from scipy.integrate import trapezoid as trapz
```
or stick the version of `scipy` to <1.14 within the `pyproject.toml` file.

xgarrido:
The fix should be backward compatible up to `scipy` 1.6.0 where the function first appears.

Fix #952

### DIFF
--- a/lib/healpy/sphtfunc.py
+++ b/lib/healpy/sphtfunc.py
@@ -1336,7 +1336,7 @@ def beam2bl(beam, theta, lmax):
     bl : array
         Beam window function b(l).
     """
-    from scipy.integrate import trapz
+    from scipy.integrate import trapezoid as trapz
 
     nx = len(theta)
     nb = len(beam)


lpsinger (commented):
What about replacing with numpy.trapz to eliminate the unnecessary dependency on scipy?

xgarrido:
The dependency to `scipy` will remain https://github.com/healpy/healpy/blob/main/lib/healpy/sphtfunc.py#L354

By the way, `numpy` also recently deprecates `trapz` in favor of `trapezoid` name https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#main-namespace

This is exactly the same piece of code https://github.com/numpy/numpy/blob/594657d38b12b1fdc7f620ed026c8df93f271cb3/numpy/lib/_function_base_impl.py#L4982C1-L5008C1 *vs.* https://github.com/scipy/scipy/blob/fffcaa8c58b6909d22063c29e4f9593059a40230/scipy/integrate/_quadrature.py#L122-L147 (I don't really understand why they don't rely to each other...)

lpsinger:
Could we do another patch release to pick this up?

PR #963, opened 2024-08-21
Add lonlat parameter in newprojplot

dreamthreebs:
Add lonlat parameter in newprojplot


zonca:
it seems like readthedocs is failing when executing the `newvisufunc_example.ipynb` notebook. Can you verify the notebook still works fine?

dreamthreebs:
the lonlat parameter in newprojplot override the lonlat function named before, so there is a name conflict. Therefore, I change the function name lonlat into thetaphi2lonlat and preserve the name rules of lonlat parameter as other function like hp.ang2pix. That should solve the problem.

PR #976, opened 2024-11-15
fix overflow in map2alm_lsq

zonca:
For large arrays `scipy.linalg.norm` overflows and gives 0 output:

```python
> norm(np.ones(int(2.5e9))*10)
0
```

`numpy`'s version of the function works fine.


mreineck:
Wow, I should have known this, since I opened the relevant `scipy` bug myself ... https://github.com/scipy/scipy/issues/18179
Perhaps I was too optimistic and expected a quicker fix.

mreineck (approved):

PR #984, opened 2025-01-09
Update newvisufunc.py

VasudevMittal:
Added extra arguments to projview functions, which help in controlling the colors of bad pixels and background

### DIFF
--- a/lib/healpy/newvisufunc.py
+++ b/lib/healpy/newvisufunc.py
@@ -76,6 +76,8 @@ def projview(
     format="%g",
     cbar=True,
     cmap="viridis",
+    badcolor="grey",
+    bgcolor="white",
     norm=None,
     norm_dict=None,
     graticule=False,
@@ -168,6 +170,10 @@ def projview(
       Display the colorbar. Default: True
     cmap : str, optional
         Specify the colormap. default: Viridis
+    badcolor: str, optional
+        Specify the color of bad pixels. default: Grey
+    bgcolor: str, optional
+        Specify the color of the background. default: White
     norm : {'hist', 'log', 'symlog', 'symlog2', None}
       Color normalization:
       hist = histogram equalized color mapping.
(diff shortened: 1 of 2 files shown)


lib/healpy/newvisufunc.py
```diff
@@ -550,7 +556,7 @@ def find_number_of_decimals(number):
                 max = m[w].max()
 
         cm, nn = get_color_table(
-            min, max, m[w], cmap=cmap, norm=norm, **norm_dict_defaults
+            min, max, m[w], cmap=cmap, norm=norm, badcolor, bgcolor, **norm_dict_defaults
```
zonca, line 559: SyntaxError: positional argument follows keyword argument

zonca (changes requested):
thanks for the contribution, please also add a short unit test, even just to check that passing badcolor doesn't cause a crash, no need for any complicated test.

lib/healpy/newvisufunc.py
VasudevMittal: Hi, I have fixed the bug. Please take a look.

lib/healpy/newvisufunc.py
```diff
@@ -76,6 +76,8 @@ def projview(
     format="%g",
     cbar=True,
     cmap="viridis",
+    badcol="grey",
```
zonca, line 79: badcolor

zonca (changes requested):

lib/healpy/newvisufunc.py
VasudevMittal: Hi. It seems like the deployment failure happened due to absence of a test file for newvisufunc.py in the test folder. Please tell me if I am supposed to do something here. 
Kind Regards 
PS: This is my first pull request for any code repo, and I am not sure if I am doing things correctly. 😅

zonca: @VasudevMittal thanks again for your contribution, I noticed there was no unit test for `newvisufunc`, I added a basic unit test, just to make sure there are no exceptions.

I also added a small test for bad and bgcolor.

zonca (approved):

zonca:
tried `gh pr merge --auto`, hoping the PR would be merged automatically when tests pass, however it just merged the PR in right away...

Issue #987, opened 2025-01-18
healpy alm2map_lsq fails with Scipy 1.15


zonca:
@mreineck @lpsinger  any suggestions on what could be causing this?

https://github.com/healpy/healpy/actions/runs/12846653655/job/35822177921#step:4:671

```
 alm = array([[0, 0, 0, ..., 0, 0, 0],
         [0, 0, 0, ..., 0, 0, 0],
         [0, 0, 0, ..., 0, 0, 0]], dtype=int8)
  
      def realalm2alm(alm):
          alm = np.array(alm).reshape((ncomp, -1))
          res = np.zeros((ncomp, (alm.shape[1] + lmax + 1) // 2), dtype=np.complex128)
          for i in range(ncomp):
              res[i, 0 : lmax + 1] = alm[i, 0 : lmax + 1]
  >           res[i, lmax + 1 :] = alm[i, lmax + 1 :].view(np.complex128) * (
                  np.sqrt(2.0) / 2
              )
  E           ValueError: could not broadcast input array from shape (260,) into shape (2080,)
```

lpsinger:
> [@mreineck](https://github.com/mreineck) [@lpsinger](https://github.com/lpsinger) any suggestions on what could be causing this?

Either a recent regression on main, or a change in scipy 1.15.1 which was released last week.

zonca:
@lpsinger you are right, no problems with scipy 1.12, I can reproduce error locally with scipy 1.15

mreineck:
Sorry, I have no idea why this is happening. Especially, looking at the failing lines it doesn't seem to be related to `scipy` at all, but rather `numpy`. I'll play around with this a bit.

zonca:
@mreineck I think it comes from `LinearOperator`

mreineck:
Oh, that would make a lot of sense - thanks for the hint!

zonca:
@mreineck any insights yet?

mreineck:
Unfortunately no, and sorry for the delay!
I tried for about an hour to set up a combinations of `scipy`, `numpy` and `healpy` versions locally which reproduces the problem, but didn't manage. Then I looked at the `scipy` ChangeLog, but couldn't identify anything indicating an interface change that would explain the behaviour. On the other hand, there were a few `scipy` issues indicating that something related to the `LinearOperator` broke in 1.15, so I hoped that things would sort themselves out ...

mreineck:
[...]

PR #988, opened 2025-01-18
Debug alm shape error in Github actions

zonca:
See #987


lpsinger:
I doubt that this is a macOS-specific issue. The macOS runners are just waaaay faster than the others, so they are the first to fail. When they fail, the other jobs are cancelled.

zonca:
fixed in #991

PR #991, opened 2025-03-20
Tentative fix for #987

mreineck:


mreineck:
I don't understand the CI failures ... Ubuntu builds seem to time out.
@zonca, @lpsinger how can we proceed with this one?

mreineck:
A bit more background on the original problem: it seems that in `scipy` 1.15, the data type for input and output arrays of a linear problem is assumed to be `np.int8`; explicitly setting the data type when constructing the `LinearOperator` seems to fix this.
I'm pretty convinced that this `scipy` change was not intentional - doing linear algebra on non-float types is just too weird. So there might be more breakage in this area of the code ...

mreineck:
No, I currently don't have the time to create a minimal reproducer. But looking at
https://github.com/scipy/scipy/issues?q=is%3Aissue%20state%3Aopen%20LinearOperator%20
makes me optimistic that the `scipy` people are aware of the problem.

zonca:
@lpsinger it seems it is giving an error with autoconfig

lpsinger:
> @lpsinger it seems it is giving an error with autoconfig

That will be fixed by https://github.com/HEASARC/cfitsio/pull/40.

lib/healpy/sphtfunc.py
```diff
@@ -397,7 +397,8 @@ def m2a2(x):
     # initial guess
     alm0 = m2a2(maps) / npix * (4 * np.pi)
     op = LinearOperator(
-        matvec=a2m2, rmatvec=m2a2, shape=(len(maps) * maps[0].size, alm0.size)
+        matvec=a2m2, rmatvec=m2a2, shape=(len(maps) * maps[0].size, alm0.size),
+        dtype=alm0.real.dtype
```
lpsinger, line 401: Would you please add a `# FIXME` comment with a link to the Scipy issue?

lpsinger (changes requested):

mreineck:
I added a comment with a link. No need for a FIXME, I think, because the explicit datatype specification doesn't hurt and can stay indefinitely.

lpsinger (approved):

PR #992, opened 2025-03-20
Set explicit Sphinx configuration for Readthedocs

zonca:


zonca:
good, readthedocs passes

PR #994, opened 2025-03-25
Build Linux aarch64 wheels natively, not under emulation

lpsinger:
This will result in faster builds.


lpsinger:
@zonca, this should fix timeouts on the Ubuntu builds like https://github.com/healpy/healpy/actions/runs/14047532377/job/39331474025?pr=991.

PR #995, opened 2025-03-25
Use Autotools and Pip packages on readthedocs, not conda

lpsinger:


lpsinger:
@zonca, this fixes the Readthedocs build.

Issue #975, opened 2024-11-08
Pixel windows at very high N_side


ajvanengelen:
Is there an "official" highest-possible N_side value that is supported by the code?  We tried making maps at N_side = 16384, but found that there was no pre-computed pixel window function available.  

Would it be possible to add these to the distribution, at higher N_side?

Thanks!

zonca:
@mreineck is it possible to create a 16384 pixel window function to add to https://github.com/healpy/healpy/tree/main/lib/healpy/data? The max available is now 8192

ajvanengelen:
That would be great -- 32k would also be helpful if possible -- thanks!

mreineck:
I think we have scripts to compute pixel windows for very high `Nside` values, but as far as I can tell, `healpy` is still limited to a maximm `Nside` of 8192:
```
ValueError: nside 16384 of map cannot be larger than MAX_NSIDE 8192
```

zonca:
@mreineck thanks,

that is related to this issue: https://github.com/healpy/healpy/issues/812

however, I think parts of healpy can be used for higher nside, so I think it would still be useful to provide pixel window functions at higher nside

mreineck:
I just did a bit of email archaeology, an I think I misremembered ... as far as I can see, I don't have the code for pixel windows, but rather the one for ring weights. @hivon, I think you have the code in question ... do you think it can provide pixel windows for Nside=16384 and 32768?

hivon:
Hi @zonca and @mreineck, I can generate such files, after digging out the proper scripts, but seamlessly supporting those large Nside across all of Healpix and healpy will require a bit of coding.

zonca:
Understood. I think that for now even just providing the files it's okay, people might need some tweaking to get those working.

hivon:
Hi @ajvanengelen, @zonca and @mreineck, the HEALPix pixel window functions for Nside=16384 and 32768 are now available in https://healpix.sourceforge.io/data/pixel_window_functions/ with self-explicit names.
At this point in time, there is NO guarantee that the whole Healpix/healpy software will correctly handle or make use of such files, which, as already pointed out, are only relevant for datasets with *very* large number of pixels.

ajvanengelen:
Thank you so much for doing that!!
Alex

zonca:
ok, I added them also here:

https://github.com/healpy/healpy-data/tree/master/pixel_window_functions

however at the moment they are not available from healpy.

PR #1005, opened 2025-04-22
Get and cache pixel window functions from Github

zonca:
- **[feat] removed pixel window functions from data folder**
- **[feat] load pixel window functions from github**
- **[feat] add datapath parameter to pixwin function for custom file location**
- **[test] test pixel windows retrieval**


zonca:
This is a followup of #975

The total space occupied by these files, once we add 16k and 32k, is more than 4 MB, so they should be hosted outside of the package.

zonca:
@lpsinger what do you think of this?

lpsinger:
What purpose does this solve? It's only 2.3 MB of data. If you want to have smaller tarballs and wheels, then what I would do is make a separate Python package that contains only the pixel data which you would only have to re-release on rare occasions. Call it `healpy-data`.

Issue #1007, opened 2025-05-16
removing small holes from dist2holes


hivon:
In Healpix/F90 `process_mask` code, which computes the angular distance of each pixel to the closest boundary of a binary mask, it is possible to ignore small holes (of user defined size) thanks to the parameters `hole_min_size` (in pixels) and `hole_min_surf_arcmin2` (in surface).
In its `healpy.dist2holes `implementation, I can't find equivalent arguments, or at least they are not documented nor explicitly available through the interface.
Would it be possible to add such features in `dist2holes` ?

mreineck:
I think that `dist2holes` in its current form doesn't support this additional functionality directly.
To keep things modular I'd suggest to have a separate function called `fill_small_holes` or similar, taking a map and either `hole_min_size` or `hole_min_surf_arcmin2`, which will simply fill small holes with any legitimate value.
Unfortunately I don't think I'll have the time to implement this anytime soon ...

zonca:
This is a test comment created by GitHub Copilot.

zonca:
A pull request is now available for this feature: https://github.com/healpy/healpy/pull/1008

This is an initial implementation created while experimenting with GPT-4.1 for code assistance. Feedback and suggestions are welcome!

zonca:
implemented in https://github.com/healpy/healpy/pull/1008

PR #1008, opened 2025-05-22
dist2holes: add support for ignoring small holes (GPT-4.1 initial implementation)

zonca:
This pull request adds support for ignoring small holes in the dist2holes function, addressing issue #1007.

Implementation plan:

1. Review Current Implementation:
   - The current dist2holes implementation does not expose parameters for ignoring small holes.
   - The Fortran90 code supports hole_min_size and hole_min_surf_arcmin2 to filter out small holes before distance computation.

2. Design API Extension:
   - Add new optional arguments to healpy.dist2holes (and the underlying C++/Cython code) for:
     - hole_min_size (int, in pixels)
     - hole_min_surf_arcmin2 (float, in arcmin2)
   - Update the docstring and documentation to describe these options.

3. Implement Hole Filtering:
   - Before computing distances, identify and fill holes in the mask smaller than the user-specified threshold.
   - This can be done by:
     - Labeling connected regions of zeros (holes) in the mask.
     - Measuring their size (in pixels and/or arcmin2).
     - Filling (setting to 1) those below the threshold.

4. Integrate with Existing Pipeline:
   - Ensure the mask with small holes filled is passed to the distance computation.
   - Maintain backward compatibility (default: do not ignore any holes).

5. Testing:
   - Add unit tests for the new feature, including edge cases (no holes, all holes, holes at threshold).
   - Validate that results match the Fortran90 behavior for equivalent parameters.

6. Documentation:
   - Update the user documentation and docstrings.
   - Add an example in the docs and/or a Jupyter notebook.

This is an initial implementation, created while experimenting with GPT-4.1 for code assistance. Feedback and suggestions are welcome.

### DIFF
--- a/src/_masktools.pyx
+++ b/src/_masktools.pyx
@@ -12,7 +12,7 @@ cdef extern from "mask_tools.h":
     cdef Healpix_Map[double] dist2holes(Healpix_Map[double] &mask,
                                         double max_distance)
 
-def dist2holes_healpy(m, maxdist=np.pi):
+def dist2holes_healpy(m, maxdist=np.pi, hole_min_size=None, hole_min_surf_arcmin2=None):
     """Computes the distance (in radians) from pixel center to center of
     closest invalid pixel up to a maximal distance.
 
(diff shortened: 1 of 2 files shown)


zonca:
✅ Updated the branch after review:

- fixed the neighbour walk so tiny holes stay isolated by switching to the RING connectivity (`fill_small_holes`) and reusing `npix2nside`.
- removed the post-processing clamp that forced `dist2holes` to zero when everything was valid; it now consistently returns `maxdist` as before.
- rewrote the new tests to construct deterministic single-pixel and extended holes using spherical coordinates so they are well separated and guard both pixel-count and area thresholds.
- expanded the `fill_small_holes` docstring with full NumPy-style parameter/return docs.

Verification: `uv run --no-editable --extra test pytest --doctest-plus --doctest-cython --pyargs healpy` and `uv run --no-editable --extra test pytest test/` both pass.

PR #1020, opened 2025-08-12
Support for adjusting latitudes automatically in lonlat2thetaphi

anaismartins:
Some context: when using `ang2pix()`, any longitudes are supported, but the same does not happen for the latitudes and an error is thrown from `check_theta_valid()` if the given latitudes are outside of the [-90, 90] range.
The conversion of longitude is more natural than that of latitude, but having the option to do it automatically for latitude too would be helpful (coming from someone who spent a long time trying to debug because I saw that the function had no problems converting longitudes but apparently it does for latitudes).


zonca:
thanks, please add unit tests both for an array and a scalar input.

zonca:
@anaismartins can you check if any of the copilot suggestions make sense and if so merge them?

zonca:
checking if CI errors are unrelated in https://github.com/healpy/healpy/pull/1024

PR #1060, opened 2025-11-11
Bundle pixel window fixtures for offline tests

zonca:


zonca:
@copilot can we use pixel window nside 4 and weights nside 32 instead?

PR #1071, opened 2025-12-16
fix(tests): make test_pixwin_local_datapath robust to network errors

zonca:
Add a retry mechanism to the download of the pixel window file to avoid test failures due to transient network errors.


zonca:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/healpy/healpy/pull/1071#pullrequestreview-3585253013)

lpsinger:
I'd really suggest using [astropy.utils.data](https://docs.astropy.org/en/stable/utils/data.html) rather than trying to get fancy.

PR #1094, opened 2026-03-15
Improve pixel lib integer type portability across platforms

pattersonjack:
Some parts of the code implicitly assume that long is 8 bytes. This is true on most Unix-like systems, but not on Windows. This change replaces those usages with explicit 64-bit types (int64_t) so the behaviour is consistent across platforms.

The issue was discovered while working on enabling Windows builds in the conda-forge healpy feedstock [PR.](https://github.com/conda-forge/healpy-feedstock/pull/84)

### DIFF
--- a/src/_healpy_pixel_lib.cc
+++ b/src/_healpy_pixel_lib.cc
@@ -113,7 +113,7 @@ template<Healpix_Ordering_Scheme scheme>static void
       if (nside!=oldnside)
         { oldnside=nside; hb.SetNside(nside, scheme); }
       try {
-        *(int64 *)op = hb.xyf2pix((int)*(long *)ip2,(int)*(long *)ip3,(int)*(long *)ip4);
+        *(int64 *)op = hb.xyf2pix((int)*(int64 *)ip2,(int)*(int64 *)ip3,(int)*(int64 *)ip4);
       } catch(PlanckError &e) {
         *(int64 *)op = -1;
       }
@@ -141,13 +141,13 @@ template<Healpix_Ordering_Scheme scheme> static void
       try {
         int x, y, f;
         hb.pix2xyf(*(int64 *)ip2, x, y, f);
-        *(long *)op1 = x;
-        *(long *)op2 = y;
-        *(long *)op3 = f;
+        *(int64 *)op1 = x;
+        *(int64 *)op2 = y;
+        *(int64 *)op3 = f;
       } catch (PlanckError & e) {
-        *(long *)op1 = -1;
-        *(long *)op2 = -1;
-        *(long *)op3 = -1;
+        *(int64 *)op1 = -1;
+        *(int64 *)op2 = -1;
+        *(int64 *)op3 = -1;
       }
     }
 }
@@ -464,10 +464,10 @@ static char pix2ang_signatures[] = {
   NPY_INT64, NPY_INT64, NPY_DOUBLE, NPY_DOUBLE
 };
 static char xyf2pix_signatures[] = {
-  NPY_INT64, NPY_LONG, NPY_LONG, NPY_LONG, NPY_INT64
+  NPY_INT64, NPY_INT64, NPY_INT64, NPY_INT64, NPY_INT64
 };
 static char pix2xyf_signatures[] = {
-  NPY_INT64, NPY_INT64, NPY_LONG, NPY_LONG, NPY_LONG
+  NPY_INT64, NPY_INT64, NPY_INT64, NPY_INT64, NPY_INT64
 };
 static char pix2vec_signatures[] = {
   NPY_INT64, NPY_INT64, NPY_DOUBLE, NPY_DOUBLE, NPY_DOUBLE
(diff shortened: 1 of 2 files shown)


zonca:
@lpsinger are you satisfied with the responses to your comments? The PR enables Windows builds in conda-forge and all CI is passing. Would be great to get this merged.

lpsinger:
> @lpsinger are you satisfied with the responses to your comments? The PR enables Windows builds in conda-forge and all CI is passing. Would be great to get this merged.

No, it doesn't add up. The C++ functions are defined as:

```c++
template<typename I> class T_Healpix_Base: public Healpix_Tables {
  /* ... */
  I xyf2pix(int ix, int iy, int face_num) const;
  void pix2xyf(I pix, int &ix, int &iy, int &face_num) const;
  /* ... */
};
```

so these arguments are defined not as `long`, but as `int`, which is a 4-byte type on both Windows and Linux (x86_64).

pattersonjack:
> > @lpsinger are you satisfied with the responses to your comments? The PR enables Windows builds in conda-forge and all CI is passing. Would be great to get this merged.
> 
> No, it doesn't add up. The C++ functions are defined as:
> 
> ```c++
> template<typename I> class T_Healpix_Base: public Healpix_Tables {
>   /* ... */
>   I xyf2pix(int ix, int iy, int face_num) const;
>   void pix2xyf(I pix, int &ix, int &iy, int &face_num) const;
>   /* ... */
> };
> ```
> 
> so these arguments are defined not as `long`, but as `int`, which is a 4-byte type on both Windows and Linux (x86_64).

Could you clarify whether you think this PR introduces a new issue, or whether you're saying the existing `NPY_LONG` signature was already wrong for these arguments?

The C++ API indeed uses `int` for `ix`, `iy`, and `face_num`, and the PR still passes `int` to it. On Linux x86_64, the old `NPY_LONG` path was already an 8-byte ufunc input narrowed to a 4-byte `int`. On Win64, `NPY_LONG` is 4 bytes, which is the mismatch this PR is trying to fix.

So I think there are two separate questions:

1. Should these ufunc arguments mirror the C++ API exactly and use `NPY_INT`?
2. Or, if the existing Linux behavior was acceptable, should the Windows behavior be made consistent with it by using `NPY_INT64`?

If we change these arguments to `NPY_INT`, the ufunc would mirror the C++ API more closely and avoid accepting values that later need to be narrowed to `int`, but default NumPy integer arrays on 64-bit platforms (`int64`) would no longer be accepted without an explicit cast to `int32`. I think that tradeoff is reasonable (although would be a breaking change).

mreineck:
I can confirm that for `ix`, `iy`, and `face_num` 4-byte integers are indeed sufficient in all situations where the pixel index fits into an 8-byte integer.
Using mixed types in the interface was probably not a great decision, I admit ... but at the time (around 2005) 64bit integers could still be quite slow on everyday hardware.

lpsinger:
@pattersonjack, I just can't tell from this PR what it is actually supposed to fix. As you can see, the arguments are narrowed down to 4 bytes anyway, so I don't see why this would fail on a platform where longs are 4 bytes.

Your PR description states that some parts of this code implicitly assume that a long is 8 bytes. This is incorrect. The code requires a long to be at least 4 bytes, which it is on both Linux and Windows.

It would help if we had Windows CI/CD so that you could demonstrate, reproducibly and in a way that I can verify, what is failing on Windows.

pattersonjack:
Thank you @mreineck for confirming that.


@lpsinger Thanks, I understand your concern now. While the C++ API accepts values that are at least 4 bytes as they are narrowed anyway, this PR addresses the NumPy boundary. 

For example, on Windows a standard call (without this patch) like:

`hp.xyf2pix(16, 8, 8, 4)`

will fail because NumPy sees the ufunc as requiring longs and cannot safely cast the default 8-byte integers down to 4-byte. Although correctly typed inputs can still be successfully passed here, existing code using healpy would need to be modified to explicitly cast values to 4-byte long on Windows.

The call above fails with:
```
  File "C:\src\healpy-feedstock\repro_xyf2pix.py", line 4, in <module>
    print(hp.xyf2pix(16, 8, 8, 4))
          ~~~~~~~~~~^^^^^^^^^^^^^
  File "C:\Users\Jack\miniconda3\envs\cf-healpy-py313-fresh\conda-bld\healpy_1779318557780\_test_env\Lib\site-packages\healpy\pixelfunc.py", line 594, in xyf2pix
    return pixlib._xyf2pix_ring(nside, x, y, face)
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
TypeError: ufunc '_xyf2pix_ring' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

Overall this results in two doctest failures (`xyf2pix` as above, and similarly one for `pix2xyf`). My PR should have included an explicit example like this initially.

The other motivation was to generally avoid platform-differing types and match the Linux behaviour.

Windows CI would definitely be ideal to catch this (as the healpy-feedstock CI does). Getting that working for wheels outside conda-forge is a longer-term goal.

lpsinger (approved):
I see. This is not about correctness, but about the convenience of being able to operate on Python integers which Numpy interprets as int64 on 64-bit systems.

PR #1106, opened 2026-06-02
Modernize some aspects of Numpy and CPython API usage

lpsinger:
- Replace soft-deprecated `PyModule_AddObject` with `PyModule_AddObjectRef`.
- Replace single-phase module initialization with two-phase module initialization, with simpler error handling.
- Remove now unnecessary workarounds for const-correctness.
- Update Numpy API target version.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,7 @@ Unreleased
 * CI: Added test workflow (`.github/workflows/tests.yml`) running the full test suite on Python 3.10–3.13 with ccache, plus an astropy pre-release job https://github.com/healpy/healpy/pull/1109
 * Fix off-by-one in ``synalm`` Cℓ boundary check that caused out-of-bounds reads when ``l == len(cl)`` https://github.com/healpy/healpy/pull/1108
 * Fix scalar boolean array indexing in ``projector.py`` projection functions (``mollview``, ``gnomview``, ``azeqview``) that caused corrupted masked output and ``"must not happen"`` assertion errors with NumPy 2.2.x + Python 3.14 https://github.com/healpy/healpy/pull/1112
+* **BREAKING**: Bump minimum NumPy version to 2.0.0 and compile extensions against the NumPy 2.0 ABI. The last stable release supporting NumPy 1.x is 1.19.0. Replaced soft-deprecated ``PyModule_AddObject`` with ``PyModule_AddObjectRef`` and converted single-phase C extension init to two-phase (``Py_mod_exec``) init. https://github.com/healpy/healpy/pull/1106
 
 Release 1.20.0b1 29 May 2026
 
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,7 +4,7 @@ sphinx:
   configuration: doc/conf.py
 
 build:
-  os: ubuntu-22.04
+  os: ubuntu-24.04
   apt_packages:
     - autoconf
     - automake
@@ -15,7 +15,7 @@ build:
     - libtool
     - pkgconf
   tools:
-    python: "3.11"
+    python: "3.12"
 
 python:
   install:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ authors = [
 ]
 license = {text = "GPL-2.0-only"}
 requires-python = ">=3.10"
-dependencies = ["numpy>=1.19", "astropy"]
+dependencies = ["numpy>=2.0.0", "astropy"]
 
 [project.optional-dependencies]
 all = ["matplotlib", "scipy"]
@@ -41,8 +41,6 @@ doc = [
     "matplotlib",
     "nbsphinx",
     "numpydoc",
-    "numpy<2",
-    "scipy<1.12",
     "sphinxnotebookgist",
     "sphinx>=6,<9",
     "sphinx-book-theme",
@@ -62,7 +60,7 @@ issues = "https://github.com/healpy/healpy/issues"
 requires = ["setuptools>=77",
             "setuptools-scm>=8.0",
             "cython>=0.16",
-            "numpy>=2.0.0rc1",
+            "numpy>=2.0.0",
             "pykg-config"]
 
 build-backend = 'setuptools.build_meta'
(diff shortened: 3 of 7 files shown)


zonca:
Rebased this onto `main` and added a one-commit fix for the `ubuntu-22.04-arm` compile failure in PR #1115:

https://github.com/healpy/healpy/pull/1115

Kept @lpsinger's original commit untouched (cherry-picked, with one trivial conflict resolved in the `xyf2pix`/`pix2xyf` signature arrays — main's #1094 portability fix had changed `NPY_LONG`→`NPY_INT64`; kept main's values and applied the PR's `static const`).

The fix replaces the inline compound literal used for the module slots

```c
(PyModuleDef_Slot []) {
      {Py_mod_exec, (void *) m_exec},
      {NULL, NULL}
}
```

with a named static array and the conventional `{0, NULL}` terminator. On the newer GCC toolchain used for the aarch64 build, `NULL` has type `std::nullptr_t`, so the sentinel's `NULL` was being passed to the `int slot` field of `PyModuleDef_Slot`, producing:

```
error: cannot convert 'std::nullptr_t' to 'int'
```

`0` is a valid `int` value for the terminating slot, so the conversion is well-formed. The `x86_64`/macOS builds already passed on #1106; this should clear the arm build that was failing.

Happy to close #1106 in favor of #1115 if that works — wanted to avoid force-pushing to @lpsinger's branch since this PR is from early June.

zonca:
Hi Leo — thanks for this, the NumPy 2 / two-phase init modernization is the right call. I dug into the two CI failures and they're both fixable with small follow-ups. Here's what's broken and what needs to change.

## 1. `ubuntu-22.04-arm` compile failure (the C++ one)

```
src/_healpy_pixel_lib.cc:653:3: error: cannot convert 'std::nullptr_t' to 'int' in initialization
```

### Why it only fails on arm

Your `moduledef` uses an **inline compound literal** for the slots:

```c
static PyModuleDef moduledef = {
  PyModuleDef_HEAD_INIT,
  "_healpy_pixel_lib",
  docstring, 0, NULL, (PyModuleDef_Slot []) {
        {Py_mod_exec, (void *) m_exec},
        {NULL, NULL}          // <-- the terminator sentinel
  }
};
```

`PyModuleDef_Slot` is `struct { int slot; void *value; }`. The terminating `{NULL, NULL}` passes `NULL` (which, on the newer GCC toolchain used for the aarch64 build, has type `std::nullptr_t`) to the `int slot` field. That implicit `nullptr` → `int` conversion is ill-formed and the build fails. The x86/macOS builds use an older toolchain where `NULL` is an integer constant, so they pass — which is exactly why this only fails on arm.

### The fix

Extract the slots into a **named static array** and use the conventional `{0, NULL}` terminator (`0` is a valid `int` for the terminating slot, so there's no `nullptr` → `int` conversion):

```c
static PyModuleDef_Slot moduledef_slots[] = {
  {Py_mod_exec, (void *) m_exec},
  {0, NULL}
};

static PyModuleDef moduledef = {
  PyModuleDef_HEAD_INIT,
  "_healpy_pixel_lib",
  docstring, 0, NULL, moduledef_slots,
  NULL, NULL, NULL
};
```

Using a named array (rather than an inline compound literal) is also the form recommended in the CPython docs for multi-phase init, and it sidesteps the compound-literal-lifetime subtleties entirely.

## 2. Read the Docs build failure (the `doc` extra)

The RTD build fails during dependency resolution, before any docs are built:

```
ERROR: Cannot install healpy[doc]==... because these package versions have conflicting dependencies.
    healpy depends on numpy>=2.0.0
    healpy[doc] depends on numpy<2
```

### Root cause

On `main`, the `doc` optional-dependency extra still pins `numpy<2` and `scipy<1.12` — a temporary workaround added in #1072, back when the docs build resolved to numpy 2.3.x and rejected extensions built against numpy 1.x. Your PR bumps the base dependency to `numpy>=2.0.0` (and retargets the extensions at the NumPy 2.0 API), but leaves the `doc` extra untouched, so `pip install .[doc]` becomes unsatisfiable.

### The fix

Update the `doc` extra to match the NumPy 2 migration:

```toml
doc = [
    "ipykernel",
    "matplotlib",
    "nbsphinx",
    "numpydoc",
    "numpy>=2.0.0",      # was numpy<2
    "scipy>=1.13.0",     # was scipy<1.12; 1.13.0 is the first release supporting NumPy 2.0
    "sphinxnotebookgist",
    "sphinx>=6,<9",
    "sphinx-book-theme",
]
```

(scipy 1.13.0 is the first release to support NumPy 2.0, which is why the floor moves to 1.13.0.)

## Summary

Both fixes are small and localized:
- **Fix 1** (arm compile): hoist the `PyModuleDef_Slot` array out of the inline compound literal, terminate with `{0, NULL}` instead of `{NULL, NULL}`.
- **Fix 2** (RTD): drop the `numpy<2`/`scipy<1.12` pins in the `doc` extra in favor of `numpy>=2.0.0`/`scipy>=1.13.0`.

Happy to test a branch if you'd like a second pair of eyes on the verification.

CHANGELOG.rst
```diff
@@ -5,6 +5,7 @@ Unreleased
 * CI: Added test workflow (`.github/workflows/tests.yml`) running the full test suite on Python 3.10–3.13 with ccache, plus an astropy pre-release job https://github.com/healpy/healpy/pull/1109
 * Fix off-by-one in ``synalm`` Cℓ boundary check that caused out-of-bounds reads when ``l == len(cl)`` https://github.com/healpy/healpy/pull/1108
 * Fix scalar boolean array indexing in ``projector.py`` projection functions (``mollview``, ``gnomview``, ``azeqview``) that caused corrupted masked output and ``"must not happen"`` assertion errors with NumPy 2.2.x + Python 3.14 https://github.com/healpy/healpy/pull/1112
+* **BREAKING**: Bump minimum NumPy version to 2.0.0 and compile extensions against the NumPy 2.0 ABI. The last stable release supporting NumPy 1.x is 1.19.0. Replaced soft-deprecated ``PyModule_AddObject`` with ``PyModule_AddObjectRef`` and converted single-phase C extension init to two-phase (``Py_mod_exec``) init. https://github.com/healpy/healpy/pull/1106
```
lpsinger, line 8: I'd remove this part:

> Replaced soft-deprecated ``PyModule_AddObject`` with ``PyModule_AddObjectRef`` and converted single-phase C extension init to two-phase (``Py_mod_exec``) init.

These are not user-visible changes.

PR #1115, opened 2026-07-09
Modernize NumPy/CPython API usage (rebase + fix #1106)

### LINKED ISSUE #1: create README for github homepage
cyrille, would you like to create  a README in the repository to present the project?
it is then shown in the github homepage, you can use material from the google code homepage.

### LINKED ISSUE #2: inconsistent shape in vec2ang()/ang2vec()
The input of vec2ang() is expected to have shape (N, 3), while its output is of shape (2, N). The axis of the number N of points should be the same at input and output.
I suggest we use the equivalent functions in rotator.py called dir2vec and vec2dir, renaming them to ang2vec and vec2dir respectively. These functions will be imported in **init**.py but stay in rotator.py.

zonca:
## Summary

This is a rebase of #1106 (by @lpsinger) onto `main`, plus fixes for the two CI failures that were blocking it. The original PR is ~6 months stale and reports `CONFLICTING`; rather than force-push to @lpsinger's branch, this keeps that PR untouched and carries the work forward here.

**Commits:**
1. `Modernize some aspects of Numpy and CPython API usage` — @lpsinger's original commit, cherry-picked. Resolved one trivial conflict in the `xyf2pix`/`pix2xyf` signature arrays (main's #1094 portability fix changed `NPY_LONG`→`NPY_INT64`; kept main's values, applied the PR's `static const`).
2. `Fix PyModuleDef slot sentinel for strict NULL toolchains` — **fix #1: the `ubuntu-22.04-arm` compile failure.**
3. `Fix RTD doc build: allow numpy 2 in doc extra` — **fix #2: the Read the Docs build failure.**

---

## Fix #1: `std::nullptr_t` → `int` on `ubuntu-22.04-arm`

PR #1106 failed on the `ubuntu-22.04-arm` (aarch64) binary builds with:

```
src/_healpy_pixel_lib.cc:653:3: error: cannot convert 'std::nullptr_t' to 'int' in initialization
```

### Root cause

The original code used an inline **compound literal** for the module slots:

```c
static PyModuleDef moduledef = {
  PyModuleDef_HEAD_INIT,
  "_healpy_pixel_lib",
  docstring, 0, NULL, (PyModuleDef_Slot []) {
        {Py_mod_exec, (void *) m_exec},
        {NULL, NULL}          // <-- the terminator sentinel
  }
};
```

`PyModuleDef_Slot` is `struct { int slot; void *value; }`. The terminating sentinel `{NULL, NULL}` passes `NULL` (which, on the newer GCC toolchain used for the aarch64 build, has type `std::nullptr_t`) to the `int slot` field. That implicit `nullptr` → `int` conversion is ill-formed and the build fails. (The x86/macOS builds use an older toolchain where `NULL` is an integer constant, so they pass — which is exactly why #1106 only failed on arm.)

### The fix

Extract the slots into a **named static array** and use the conventional `{0, NULL}` terminator (where `0` is a valid `int` value for the terminating slot, so there is no `nullptr` → `int` conversion):

```c
static PyModuleDef_Slot moduledef_slots[] = {
  {Py_mod_exec, (void *) m_exec},
  {0, NULL}
};

static PyModuleDef moduledef = {
  PyModuleDef_HEAD_INIT,
  "_healpy_pixel_lib",
  docstring, 0, NULL, moduledef_slots,
  NULL, NULL, NULL
};
```

Using a named array (rather than an inline compound literal) is also the form recommended in the CPython docs for multi-phase init, and it sidesteps the compound-literal-lifetime subtleties entirely.

**Verified locally** against NumPy 2.4.6 / Python 3.13: full pixel-lib ufunc surface works, `map2alm`/`alm2map`/`anafast` roundtrips work, and `pytest test/test_pixelfunc.py test/test_pixel_windows.py test/test_fitsfunc.py` → **65 passed**.

**CI result:** `Build binary packages (ubuntu-22.04-arm)` now passes (the one that failed on #1106).

---

## Fix #2: Read the Docs build failure

The Read the Docs build was failing during dependency resolution (~67s, before any docs were built):

```
ERROR: Cannot install healpy[doc]==... because these package versions have conflicting dependencies.
    healpy depends on numpy>=2.0.0
    healpy[doc] depends on numpy<2; extra == "doc"
```

### Root cause

PR #1106 bumps the base `healpy` dependency to `numpy>=2.0.0` and retargets the compiled extensions at the NumPy 2.0 C API. But the `doc` optional-dependency extra still pinned `numpy<2` and `scipy<1.12` — a temporary workaround added in #1072, back when the docs build resolved to numpy 2.3.x and rejected extensions built against numpy 1.x. With `numpy>=2.0.0` in the base dependencies, `pip install .[doc]` became unsatisfiable.

### The fix

Update the `doc` extra to match the NumPy 2 migration:

| Before | After | Why |
|---|---|---|
| `numpy<2` | `numpy>=2.0.0` | extensions now target the NumPy 2.0 API |
| `scipy<1.12` | `scipy>=1.13.0` | scipy 1.13.0 is the first release to support NumPy 2.0 |

**CI result:** RTD build now succeeds (462s, full docs build) → https://healpy--1115.org.readthedocs.build/en/1115/

---

## Notes for review

- The `numpy>=2.0.0` runtime bump is a real (intended) dependency change from the original PR — flagging it here so it's visible. The build already required numpy 2.0 (`numpy>=2.0.0rc1` build-dep), so the runtime floor is being brought in line with what the compiled extensions actually target.
- Fix #2 (`doc` extra) is a small but important follow-up that #1106 needed once the numpy 2 migration landed — without it, the docs build is unsatisfiable.
- Once this merges, #1106 can be closed. I've commented there pointing here.


zonca:
## Update: RTD failure root-caused and fixed

The Read the Docs build was failing during dependency resolution (~67s, before any docs were built):

```
ERROR: Cannot install healpy[doc]==... because these package versions have conflicting dependencies.
    healpy depends on numpy>=2.0.0
    healpy[doc] depends on numpy<2; extra == "doc"
```

**Root cause:** PR #1106 bumps the base `healpy` dependency to `numpy>=2.0.0` and retargets the compiled extensions at the NumPy 2.0 C API. But the `doc` optional-dependency extra still pinned `numpy<2` and `scipy<1.12` — a temporary workaround added in #1072, back when the docs build resolved to numpy 2.3.x and rejected extensions built against numpy 1.x. With `numpy>=2.0.0` in the base dependencies, `pip install .[doc]` became unsatisfiable.

**Fix (commit `9c69c07`):** update the `doc` extra to match the NumPy 2 migration:

| Before | After | Why |
|---|---|---|
| `numpy<2` | `numpy>=2.0.0` | extensions now target the NumPy 2.0 API |
| `scipy<1.12` | `scipy>=1.13.0` | scipy 1.13.0 is the first release to support NumPy 2.0 |

**Result:** RTD build #33529013 now succeeds (462s, full docs build) — https://healpy--1115.org.readthedocs.build/en/1115/

All other checks remain green (Python 3.10–3.13, ubuntu-22.04-arm, source package, uv-editable-install, block-default-branch-prs).

lpsinger:
Would you please just review my PR rather than having AI rewrite it?

zonca:
No sorry I don't know enough about it to review

zonca:
Closing this in favor of reviewing #1106 directly — @lpsinger asked that we review his original PR rather than carry a parallel rebase. The two CI failures on #1106 (the `ubuntu-22.04-arm` `std::nullptr_t → int` compile error and the Read the Docs build failure) are documented in a review over there so Leo can apply the fixes himself.

PR #1120, opened 2026-07-25
Skip linking cfitsio and healpix_cxx programs for Pyodide builds

hoodmane:
They fail to link with duplicate symbols and are unneeded.

Upstreaming this patch: https://github.com/pyodide/pyodide-recipes/blob/main/packages/healpy/patches/0001-build-no-binaries.patch

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,6 @@
 Unreleased
-|
+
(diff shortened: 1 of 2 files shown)


lpsinger (changes requested):
What are the advantages and disadvantages of setting this here, or in the cibuildwheel configuration?

hoodmane:
I do not believe it is possible to do this in the cibuildwheel config alone. I think it requires a change to setup.py.

lpsinger:
OK. Would you please configure cibuildwheel to generate emscripten wheels, so that we have a test for this platform?

PR #1122, opened 2026-09-01
Do not fold the caller's arrays in place in lonlat2thetaphi

dylanpulver:
`lonlat2thetaphi(..., latauto=True)` folds the latitudes in place on `np.asarray(lat)`, which does not copy an ndarray. The caller's arrays come back modified, and read-only input raises:

```python
>>> lat = np.array([-100., -90., 0., 90., 100.]); lon = np.array([0., 60., 90., 0., 180.])
>>> hp.ang2pix(16, lon, lat, lonlat=True, latauto=True, latbounce=False)
>>> lat, lon
(array([-80., -90.,   0.,  90.,  80.]), array([180.,  60.,  90.,   0., 360.]))

>>> ro = np.array([100.]); ro.flags.writeable = False
>>> hp.ang2pix(16, np.array([0.]), ro, lonlat=True, latauto=True)
ValueError: assignment destination is read-only
```

Same class as gh-262 (`ud_grade modifies input array`). Fix is to fold copies; the returned values are unchanged for every input.

Why the existing test cannot see it: `test_latauto_and_latbounce` passes Python lists, and `np.asarray` always copies a list. It also reuses one `lat`/`lon` pair across three successive calls, which only works because of that copy. Swap those lists for float arrays and the existing third assertion fails on unmodified main — `phi` comes back `[0, 60, 90, 0, 180]` deg instead of `[180, 60, 90, 0, 360]`.

Ran (same venv, Python 3.13, macOS arm64): `pytest test/` 341 passed / 1 skipped before, 343 passed / 1 skipped after, no failures either way; `pytest --doctest-plus --pyargs healpy` 46 passed. Mutants, both new tests fail on each — (A) revert to `np.asarray`; (B) `np.asarray(lat, dtype=np.float64)`, the plausible "make it an array" fix, which does not copy a float64 input; (C) copy `lat` only, leaving `lon` aliased.

Not tested here: the `--doctest-cython` leg, the wheel builds, and any platform other than macOS arm64 / Python 3.13.

Separate defect, deliberately left out of this PR: the fold is one pass per sign and `lat[lat > 90]` runs before `lat[lat < -90]`, so `+280` converges to `-80` but `-280` lands on `+100`, outside the `[-90, 90]` the docstring promises — `lonlat2thetaphi([10.], [-280.], latauto=True)` returns `theta = -0.1745 rad`. Happy to open a follow-up.

Implementation, tests and this description were drafted with Claude Opus 5.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,8 @@
 Unreleased
 |
 
+* Fixed ``lonlat2thetaphi(..., latauto=True)`` (and therefore ``ang2pix(..., latauto=True)``) modifying the caller's ``lat``/``lon`` arrays in place and raising ``ValueError`` on read-only input https://github.com/healpy/healpy/pull/1122
+
 Release 1.20.0 22 Jul 2026
 
 * Fixed ``test_rotate_alm2``: the ``a_lm`` initialization now includes the ``m == ell`` diagonal, matching the Fortran reference (the test previously compared the reference to itself) https://github.com/healpy/healpy/pull/1118
--- a/lib/healpy/pixelfunc.py
+++ b/lib/healpy/pixelfunc.py
@@ -179,14 +179,16 @@ def lonlat2thetaphi(lon, lat, latauto=False, latbounce=True):
       The co-latitude and longitude in radians
     """
     if latauto:
-        lat = np.asarray(lat)
+        # np.asarray does not copy an ndarray input, so folding in place would
+        # modify the caller's arrays and fail outright on read-only input.
+        lat = np.array(lat, copy=True)
         if not latbounce:
-            lon = np.asarray(lon)
+            lon = np.array(lon, copy=True)
             lon[lat > 90] = lon[lat > 90] + 180
             lon[lat < -90] = lon[lat < -90] + 180
         lat[lat > 90] = 180 - lat[lat > 90]
         lat[lat < -90] = -(180 + lat[lat < -90])
-          
+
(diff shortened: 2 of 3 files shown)


zonca:
Hi @dylanpulver, thanks for finding this issue and for the careful regression tests. In particular, catching that `np.asarray()` can alias the caller's arrays was very useful.

I opened #1123 as an alternative implementation that addresses the same bug while avoiding full-size copies of the longitude and latitude arrays.

Instead, #1123 keeps the input `lat` array read-only and performs the folding directly in the newly allocated floating-point `theta` and `phi` output buffers. This has a few advantages:

- it preserves the non-mutation and read-only input guarantees identified here;
- it avoids allocating full copies of `lat` and `lon`, which substantially reduces peak memory for large coordinate arrays;
- the original latitude values remain available for the pole-crossing test when `latbounce=False`;
- longitude is converted to floating point before adding 180 degrees, avoiding overflow with narrow integer inputs;
- it also extends the regression coverage to strided/non-contiguous inputs, scalar behavior, output aliasing, and narrow integer dtypes.

For a test with two million float64 longitude/latitude values and `latbounce=False`, peak result/temporary allocation was about 61 MiB with the copy-based implementation versus 34 MiB with the output-buffer implementation.

I'm going to close this PR in favor of #1123 so we have a single implementation to review. I'd very much appreciate your feedback on #1123, especially if you see anything from your original implementation or tests that should be carried over or handled differently.

Thanks again for identifying and documenting the bug.

